PR #34, opened 2015-03-19
fixed bug in network not using insecure connection

FrankCRoth:
Hi, adding network did not have the create_vdc_network did not have verify=self.verify in the post and therefore failed to add network in insecure mode.


pacogomez:
thanks!

PR #36, opened 2015-03-30
Add new properties to vms_network_info

kostya13:
Add new properties:
is_primary
allocation_mode

 to vms_network_info result list


pacogomez:
thanks

PR #39, opened 2015-04-08
Add checks: list has elements.

kostya13:
Before fix if self.response.history is empty list, the  statement:
self.response.history[-1]
raise exeption: IndexError: list index out of range


pacogomez:
thanks!

PR #43, opened 2015-04-20
Use requests' HTTPBasicAuth for authentication

voyageur:
Instead of manually creating the basic auth header, this uses requests' module
I had to track that one down as I had a netrc file in my profile, and this overrides manually set headers (getting me 401 answers hard to understand at first) 

(tested on vloud director only, but the logic is the same for vcloud air)

### DIFF
--- a/pyvcloud/vcloudair.py
+++ b/pyvcloud/vcloudair.py
@@ -18,7 +18,6 @@
 #todo: upload/download ovf to/from catalog
 
 import time
-import base64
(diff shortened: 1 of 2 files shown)


voyageur:
I rebased to handle the new Http class (and the new login case for vcloud air)

pacogomez:
Thanks, this is a good improvement. I need to test it with vCloud Air, but I'll add it.

PR #44, opened 2015-04-21
Clean requirements

voyageur:
Some dependencies are only useful for vca-cli, this cleans the requirements.txt (and an unused import)
PyYAML, tabulate and xmltodict should be moved to vca-cli dependencies
I could not find any use in vca for wsgiref but maybe missed it


pacogomez:
Thanks!

PR #49, opened 2015-05-19
Add Task lists to disk object in  _parse_disk 

kostya13:


kostya13:
Fix two bugs: 
- Extra argument in requests.delete call
- Add Task lists to disk object in  _parse_disk (because Disk object  contains Task list in response)

pacogomez:
Thanks

PR #50, opened 2015-05-19
Fix logging logic to handle case where logging is not enabled

nmishkin:


pacogomez:
thanks

pacogomez:
tagged as 13c2 and uploaded to pypi

PR #52, opened 2015-06-04
Use case-insensitive comparison for rule parameters in function delete_fw_rule

kostya13:
For more convenient using function delete_fw_rule, user can set value of parameters ("any", "interal", "extermal") in any case.


pacogomez:
thanks

PR #56, opened 2015-06-11
SCOR-5 Change the Score Rest server to allow ids without Org-ids

kostya13:
Send archive using Streaming Uploads


pacogomez:
thanks,
in the future, we need to add test cases for all commits

Issue #41, opened 2015-04-09
Create is failing with vmname


suthanvs:
When I try to create vapp with vmname , its failing with error as "BAD REQUEST" , Invalid content was found starting with element 'SourcedItem' .

lasko:
Having this same issue.

can't create the vApp, <?xml version="1.0" encoding="UTF-8"?>
<Error xmlns="http://www.vmware.com/vcloud/v1.5" minorErrorCode="BAD_REQUEST" message="Bad request
 - Unexpected JAXB Exception
 - cvc-complex-type.2.4.a: Invalid content was found starting with element 'SourcedItem'. One of '{&quot;http://www.vmware.com/vcloud/v1.5&quot;:IsSourceDelete, &quot;http://www.vmware.com/vcloud/v1.5&quot;:SourcedVmInstantiationParams, &quot;http://www.vmware.com/vcloud/v1.5&quot;:AllEULAsAccepted}' is expected." majorErrorCode="400" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" ......................................

suthanvs:
Yeah lasko , I have fixed it. Its basically because of payload while making a request . I have fixed this issue on my fork but couldn't push it as I don't have permissions

pacogomez:
I'll review the pull request and will integrated it into the master branch. Thanks

lasko:
Whats the status of this being integrated into the master branch?

ACCORDS-PLATFORM:
Hello, 
I notified the same issue when reconfiguring CPUs and MEMORY

###REQUEST:
 POST  https://vcloudcell.prologue.prl/api/vdc/351bdd7f-8197-4c98-8c90-35b154aee3ab/action/instantiateVAppTemplate

Authorization: Basic YWRtaW5wcmxAcHJvbG9ndWU6UHIwbDBndWU6MjAxNA==
Accept: application/*+xml;version=5.5
Content-Type: application/vnd.vmware.vcloud.instantiateVAppTemplateParams+xml

BODY
<?xml version="1.0" encoding="UTF-8"?><InstantiateVAppTemplateParams xmlns="http://www.vmware.com/vcloud/v1.5" xmlns:ovf="http://schemas.dmtf.org/ovf/envelope/1"
[...]

PR #59, opened 2015-06-17
Fixes Issue #41 for users who are using vCloud Director API version 5.5

lasko:
For users who are on API version 5.5 there is an issue when using create_vapp and having a vm_name set.

The 5.5 API doesn't like the ```<SourcedItem>``` tag with VmGeneralParams and as a result will respond with an Error as it is not part of the 5.5 API

The resolution is to set the SourcedItem = None and not setting any vcloudType.VmGeneralParamsType parameters since the VmGeneralParamsType was not added to the API until 5.6  (Reference: http://pubs.vmware.com/vca/index.jsp?topic=%2Fcom.vmware.vcloud.api.reference.doc_56%2Fdiff%2Flanding-all_types_diff.html)

And since we want to change a name, we can set the templateParams.set_name function to be the vm_name

Its not the perfect solution but it will help those who are on 5.5

### DIFF
--- a/pyvcloud/schema/vcd/v1_5/schemas/vcloud/vcloudType.py
+++ b/pyvcloud/schema/vcd/v1_5/schemas/vcloud/vcloudType.py
@@ -8823,9 +8823,8 @@ class InstantiateVAppParamsType(VAppCreationParamsType):
             outfile.write('<%sIsSourceDelete>%s</%sIsSourceDelete>%s' % (namespace_, self.gds_format_boolean(self.IsSourceDelete, input_name='IsSourceDelete'), namespace_, eol_))
         for SourcedVmInstantiationParams_ in self.SourcedVmInstantiationParams:
             SourcedVmInstantiationParams_.export(outfile, level, namespace_, name_='SourcedVmInstantiationParams', pretty_print=pretty_print)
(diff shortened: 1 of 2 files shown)


pacogomez:
This sounds good, the only concern is that vcloudType.py is automatically generated from xml schema file. Is there any way to avoid manually changing this file? if we generate it in the future, the manual changes will be gone.

lasko:
I can't really think of any other way to do this without changing that file. The problem here is that vcloudType with my change will Always assume that a SourceItem will exist, and unfortunately in versions older than 5.6 -- this is just not the case. There are a few methods here that I think could use to resolve this.

 1.) Do what I've done and just change the vcloudType file.
 2.) Fix the XML schema file so it generates as I've provided.
 3.) Somehow overwrite the exportChildren method from within pyvcloudair in the instance that self.version == '5.5'

The last option is very hackish and I think would really ugly up the code very quickly.

lasko:
Actually I may have another option that doesn't involve any of those options..

lasko:
Setting it to just 'pass' is actually a better option than setting it to None. I hadn't thought about this previously but its simpler and doesn't mess with the vcloudType schema

pacogomez:
thanks

PR #60, opened 2015-06-18
Added real file operations to the upload_media method

piraz:
I've just added real file operations to the upload_media method. Now if file doesn't exists than the methods will fail and file size is calculated and added to the media creation request body.


pacogomez:
See commit 1cb86e7516d29ec51ef0d8cf011b9def88e57962 for the complete implementation of media uploads. It has been tested with vCloud Air and should also work with vCD Standalone. I'm closing this pull request, but thanks for contributing.

Issue #62, opened 2015-06-22
Version issues with 5.5 and older


lasko:
The latest release of the code creates the same issues as #41 as a result of the fix in #61 

The fix should have been something like this:
```
+
+        if vm_name or vm_cpus or vm_memory:
+            params = vcloudType.SourcedCompositionItemParamType()
+            if ((self.version == "1.0") or (self.version == "1.5") 
+                    or (self.version == "5.1") or (self.version == "5.5")):
+                pass
+            else:
```

I'll put in a pull request for this but I'm not sure how this affects vca-cli as I'm not using it in my environment.

pacogomez:
fixed in current master, thanks for reporting

Issue #61, opened 2015-06-19
Creating Vapp with vm name have changed vapp name


kastesh:
I've used vca-cli, that uses pyvcloud.
I've created a vApp  from a template;  API version is  5.5.

This request is doing great:
```
vca vapp create --vapp test_vapp --catalog 'vms' --template 'template_name' --network 'network_name' --mode POOL --vdc vdc_name
```
But if I've added --vm vm_name, I get the error:
```
vca  vapp create --vapp test_vapp --vm vm_name --catalog 'vms' --template 'template_name' --network 'network_name' --mode POOL --vdc vdc_name
                   Traceback (most recent call last):
  File "/bin/vca", line 9, in <module>
    load_entry_point('vca-cli==12', 'console_scripts', 'vca')()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/vca_cli/vca_cli.py", line 705, in vapp
    vca.vcloud_session.get_vcloud_headers())
  File "/usr/lib/python2.7/site-packages/vca_cli/vca_cli.py", line 2460, in display_progress
    sys.stdout.write("\r" + task_table(response.content) + '\n')
UnboundLocalError: local variable 'response' referenced before assignment
```
It is happen because, there is next condition in the pyvcloud code:
https://github.com/vmware/pyvcloud/blob/master/pyvcloud/vcloudair.py#L381
```
           if vm_name:
                if self.version == "5.5":
                   templateParams.set_name(vm_name)
                else:
                   gen_params = vcloudType.VmGeneralParamsType()
                   gen_params.set_Name(vm_name)
                   params.set_VmGeneralParams(gen_params)
```
This part is redefined name for created vApp. 
The situtation seems very odd, the vApp is created with name vm_name, not with name test_vapp.

pacogomez:
[...]

PR #63, opened 2015-06-22
Fixes issues #62

lasko:
removes the SourcedItem tag being implemented in versions == 5.5 and older.

Should resolve issues #61, and #41


pacogomez:
this should fix the issue, tested with a 5.5 VCD instance, thanks

PR #69, opened 2015-07-15
Change return code to 200 because score service only proxy request

kostya13:


GlaciErrDev:
LGTM +1

denismakogon:
LGTM +1

yoramw:
Why not throw an exception only if the return code of the HTTP is not 20x where x is [0-9]?

PR #71, opened 2015-07-17
Add parameter 'interface' to NAT rule operations

kostya13:


pyvcloud/gateway.py
```diff
@@ -157,21 +165,21 @@ def add_nat_rule(self, rule_type, original_ip, original_port, translated_ip, tra
...
+    def del_nat_rule(self, rule_type, original_ip, original_port, translated_ip, translated_port, protocol, interface=None):
+        gatewayInterfaces = self._select_gateway_interface(interface)
         if len(gatewayInterfaces) == 0:
             return False
         gatewayInterface = gatewayInterfaces[0]
         natRules = self.get_nat_rules()
         newRules = []
-
+        if interface:
+            interface = interface.lower()
```
pacogomez: If interface is None, it should assume is the first uplink interface by default (for example 'd3p4v54-ext'), but here the interface will still be None, even if gatewayInterface is the uplink interface returned by _select_gateway_interface()

Another issue is that if interface value is provided, _select_gateway_interface will return an interface only if the name matches an internal interface, if the name refers to an uplink interface, _select_gateway_interface will not find it.

_select_gateway_interface, del_nat_rule and add_nat_rule need to be modified to meet this requirement

PR #73, opened 2015-07-31
Fix url for getting plans and instances

kostya13:


pyvcloud/vcloudair.py
```diff
@@ -245,7 +245,8 @@ def get_service_groups(self):
         headers['Accept'] = "application/json;version=%s;class=com.vmware.vchs.billing.serviceGroups" % self.version
         self.response = Http.get(self.host + "/api/billing/service-groups", headers=headers, verify=self.verify, logger=self.logger)
         if self.response.history and self.response.history[-1]:
```
pacogomez: I believe this workaround should not be necessary anymore, now that we are using 'vca.vmware.com' as the default endpoint for service type 'vca' (on-demand)

pacogomez:
please note we are using new 'ids' for the service types, and keeping the old values for compatibility reasons: on demand->vca, subscription->vchs, vcd->standalone
https://github.com/vmware/pyvcloud/blob/master/pyvcloud/vcloudair.py#L56
the default url has changed for vca: from iam.vchs.vmware.com to vca.vmware.com

PR #74, opened 2015-08-06
Revert casting from boolean to strings

denismakogon:
Revert casting from boolean to strings.
Score expects to have booleans instead of sting during install executions.


yoramw:
+1

PR #75, opened 2015-08-11
Improve Score user experience

denismakogon:
NOTE: this change is a dependency for vca-cli latest code

PR #76, opened 2015-08-11
SCOR-160: fix cancel

0lvin:


pacogomez:
I've merged PR #76  and #75, let me know if you see any issue

PR #79, opened 2015-08-13
Make exception block syntax py3-compatible

denismakogon:
Make exception block syntax py3-compatible


pacogomez:
These source files are generated from XSDs files automatically, using generateDS, https://pythonhosted.org/generateDS. It would be good if there is an option in generateDS to generate python3 compatible code, if not, are these changed done manually or do you have a script to bulk change the code?

denismakogon:
@pacogomez i did changes with small script (handmade).

Issue #68, opened 2015-07-13
Errors in automatic generated files


kostya13:
In several files in dir  pyvcloud/schema/vcd/v1_5/schemas/vcloud (for example networkType.py)
I found error in GDSClassesMapping dictonary.
The node's root tag name and corresponding class are swapped.
I found two mismatches:
'AllocatedIpAddresses': IpAddressesType
'IpAddress': AllocatedIpAddressType
Therefore if I try parse string with root node AllocatedIpAddresses, the parser try use build function from IpAddressesType class.

denismakogon:
@kostya13 fix proposed, https://github.com/vmware/pyvcloud/pull/80

hodgesrm:
These files no longer exist.  Closing as obsolete.

PR #80, opened 2015-08-13
Issue #68: Fix inconsistency

denismakogon:


pyvcloud/schema/vcd/v1_5/schemas/vcloud/networkType.py
```diff
@@ -13860,12 +13860,12 @@ def buildChildren(self, child_, node, nodeName_, fromsubclass_=False):
     'ParentNetwork': ReferenceType,
     'References': ReferencesType,
     'SubAllocations': SubAllocationsType,
-    'AllocatedIpAddresses': IpAddressesType,
+    'AllocatedIpAddresses': AllocatedIpAddressType,
```
pacogomez: In the original XSD, AllocatedIPAddresses is defined twice, so I think it picked up the last definition (IpAddressesType). If it needs to be changed, it should be to 'AllocatedIpAddressesType, not to AllocatedIpAdressType.
Once this is changed, we need to test if other parts of the code expect the type to be IpAddressesType, as originally auto-generated

pyvcloud/schema/vcd/v1_5/schemas/vcloud/networkType.py
```diff
@@ -13860,12 +13860,12 @@ def buildChildren(self, child_, node, nodeName_, fromsubclass_=False):
...
     'PeerSubnet': IpsecVpnSubnetType,
     'ServiceProfile': LBVirtualServerServiceProfileType,
     'StaticRoutingService': StaticRoutingServiceType,
     'VirtualServer': LoadBalancerVirtualServerType,
-    'IpAddress': AllocatedIpAddressType,
+    'IpAddress': IpAddressesType,
```
pacogomez: IpAddress is also defined as IpAddress and as AllocatedIpAddressType. If we change to IpAddressType (Singular) we need to validate that other parts of the code don't break.

PR #93, opened 2016-05-11
Fix get_service_type cannot detect standalone vCD

higebu:
`get_service_type` cannot detect standalone vCD, because vCD 8.0 returns `403 Forbidden` error when requesting to `https://<vcd_host>/api/iam/login`.


pacogomez:
Duplicate of PR #92 , already merged

PR #100, opened 2017-01-03
Added options to select storage profile, bus type and add comment when creating an independent disk

kaneda-fr:
Since API version 5.6 allows a single virtual machine (VM) to access different tiers of storage, this patch will allow for the selection of a storage profile when creating an independent disk.

I have modified add_disk to accept the following additional parameters :

- bus_type to define a specific Disk.busType
- bus_sub_type to define a specific Disk.busSubType
- comment that will be visible in the disk details
- storage_profile_name to define a specific storage profile for that disk (must be in the allowed storage profile defined for the vDC)

### DIFF
--- a/pyvcloud/vapp.py
+++ b/pyvcloud/vapp.py
@@ -442,7 +442,7 @@ class VAPP(object):
                 return taskType.parseString(self.response.content, True)
 
 
-    def attach_disk_to_vm(self, vm_name, disk_ref):
+    def attach_disk_to_vm(self, vm_name, disk_ref, bus_number=None, unit_number=None):
         """
         Attach a disk volume to a VM.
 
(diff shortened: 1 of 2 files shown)


pacogomez:
I will review and provide feedback

pacogomez:
thanks

cpumanaz:
Can you modify this one to create/return a task?  There is time taken to create an independent disk, so immediately attaching the disk in a subsequent step fails.

Can loop through getting disk status , but centralizing this with existing task code would be useful.

kaneda-fr:
shoud not be a problem, will submit PR

kaneda-fr:
cpumanaz, you can already track the task, have you tried:
task = the_vapp.attach_disk_to_vm(vm_name, disk_href)

You have a good example of using this in the vca_cli code.

Harish0596:
Hi,

Is the above mentioned code to select storage profile, bus type and add comment when creating an independent disk got merged ? I am not able to in current pyvcloud version

shashim22:
Hi,

This was done long time ago. The modified file doesn't exist now, there might have been significant restructuring of the code. 

-Shashi

Harish0596:
Hi Shashi,

I couldn't find any significant restructuring of code it was totally removed, it will a great help if you help me out in finding that.

Thanks

kaneda-fr:
Hi Harish, 

This patch was done on a different code base (older version) and now that I no longer have access to vCloud Director, I will not be able to recreate it. 

The patch was for attach_disk_to_vm() function and I can see the commits are still visibles, so you should be able to adapt eventually.

Rgds,
Seb

PR #106, opened 2017-07-24
Tests code cleanup

blackms:
Cleanup and fix travis-ci tests.


pacogomez:
we are currently not using travis for any functional tests, we are using .gitlab for tests with an internal vCloud Director instance. I might remove the script section of the .travis.yml, otherwise travis will fail

PR #107, opened 2017-07-28
added bravado client calls for CLUSTER and nose tests for it

auppunda:
Signed-off-by: auppunda <auppunda@github.com>


auppunda:
This pull request uses the bravado client library to make calls to the container-service-extension api
Also file added to test the cluster class

more tests to be added later

PR #116, opened 2017-10-31
Independent Disks

pacogomez:


pacogomez:
yes, we'll add iops support.

PR #122, opened 2017-11-14
Curt branch

ckolovson:
Added update_catalog function and unit test.

### DIFF
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -107,6 +107,7 @@ class Org(object):
                 return self.client.get_resource(link.href)
         raise Exception('Catalog not found.')
 
+    #update_catalog updates the catalog name and/or the catalog's description
     def update_catalog(self, old_catalog_name, new_catalog_name, description):
         org = self.client.get_resource(self.href)
         links = get_links(org,
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -117,7 +117,7 @@ class Org(object):
         """  # NOQA
         if self.resource is None:
             self.resource = self.client.get_resource(self.href)
-        org = self.client.get_resource(self.href)
+        org = self.resource
         links = get_links(org,
                           rel=RelationType.DOWN,
                           media_type=EntityType.CATALOG.value)
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -30,7 +30,6 @@ import tarfile
 import tempfile
 import time
 import traceback
-import pprint
 
 
 DEFAULT_CHUNK_SIZE = 1024*1024
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -115,6 +115,8 @@ class Org(object):
         :param description: (str): The new description of the catalog.
         :return:  A :class:`lxml.objectify.StringElement` object describing the updated catalog.
         """
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
         org = self.client.get_resource(self.href)
         links = get_links(org,
                           rel=RelationType.DOWN,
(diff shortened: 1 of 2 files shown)


pyvcloud/vcd/org.py
```diff
@@ -107,6 +107,28 @@ def get_catalog(self, name):
                 return self.client.get_resource(link.href)
         raise Exception('Catalog not found.')
 
+    def update_catalog(self, old_catalog_name, new_catalog_name, description):
```
anusuyar: Can you add docstring for this method.

ckolovson: I added a comment before the method. Is that what you meant by "docstring"?

anusuyar: No. it has to be after the method name. You can follow the docstring format present in pyvcloud/vdc.py

ckolovson: OK, done.

pyvcloud/vcd/org.py
```diff
@@ -107,6 +107,35 @@ def get_catalog(self, name):
...
+        """
+        Update the name and/or description of a catalog.
+        :param old_catalog_name: (str): The current name of the catalog.
+        :param new_catalog_name: (str): The new name of the catalog.
+        :param description: (str): The new description of the catalog.
+        :return:  A :class:`lxml.objectify.StringElement` object describing the updated catalog.
+        """
+        org = self.client.get_resource(self.href)
```
rocknes: if self.resource is None:
             self.resource = self.client.get_resource(self.href)

Will save an extra ReST call to vCD if Org object is already initialized.

pyvcloud/vcd/org.py
```diff
@@ -107,6 +107,35 @@ def get_catalog(self, name):
...
+        links = get_links(org,
+                          rel=RelationType.DOWN,
+                          media_type=EntityType.CATALOG.value)
+        for link in links:
+            if old_catalog_name == link.name:
+                catalog = self.client.get_resource(link.href)
+                href = catalog.get('href')
+                admin_href = href.replace('/api/catalog/', '/api/admin/catalog/')
+                catalog = self.client.get_resource(admin_href)
```
rocknes: Probably it's a good idea to highlight that the new catalog object we get as a result of the admin call is not the same as the old catalog object, but rather a admin view of the catalog. I think we should use a new variable say adminViewOfCatalog.

tests/vcd_catalog_update.py
```diff
@@ -0,0 +1,25 @@
...
+    def test_create_catalog(self):
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
+        catalog = org.create_catalog(self.config['vcd']['catalog'], 'test catalog')
+        assert self.config['vcd']['catalog'] == catalog.get('name')
+
+    def test_update_catalog(self):
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
+        catalog = org.update_catalog(self.config['vcd']['catalog'], self.config['vcd']['new_name'], self.config['vcd']['new_desc'])
```
rocknes: Line too long add a #  NOQA at the end of the line.

anusuyar: You can probably make the line into 2. I think the line length limit is 80 chars.

anusuyar:
Reviewed 1 of 2 files at r1, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/122)*

ckolovson:
In update_catalog, there is this code sequence:
---
if self.resource is None:
    self.resource = self.client.get_resource(self.href)
org = self.client.get_resource(self.href)
---
Should the third line be replaced by:
org = self.resource
?

PR #123, opened 2017-11-14
storage profile support

pacogomez:


pyvcloud/vcd/utils.py
```diff
@@ -174,6 +174,9 @@ def vapp_to_dict(vapp, metadata=None):
             env = vm.xpath('//ovfenv:Environment', namespaces=NSMAP)
             if len(env) > 0:
                 result['%s: %s' % (k, 'moid')] = env[0].get('{http://www.vmware.com/schema/ovfenv}vCenterId')  # NOQA
+            if hasattr(vm, 'StorageProfile'):
+                result['%s: %s' % (k, 'storage-profile')] = \
```
rocknes: It will be good if we use a bit more descriptive variable names to store vm names than just 'k'.

PR #126, opened 2017-11-15
VCDA-197 : [Independent Disks] Update an existing independent disk size,iops and storage profile

sompa:


pyvcloud/vcd/vdc.py
```diff
@@ -391,21 +391,32 @@ def add_disk(self,
...
+                    storage_profile_name=None,
+                    iops=None,
                     disk_id=None):
         """
         Update an existing independent disk.
         :param name: (str): The existing name of the Disk.
-        :param new_name: (str): The new name for the Disk.
         :param size: (str): The size of the new disk in MB.
+        :param new_name: (str): The new name for the Disk.
+        :param iops: (str): The iops for the disk.
```
rocknes: nit : new iops value

pyvcloud/vcd/vdc.py
```diff
@@ -415,6 +426,12 @@ def update_disk(self, name, size, new_name=None, description=None,
         else:
             disk = self.get_disk(name)
 
+        if storage_profile_name is not None:
+            storage_profile = self.get_storage_profile(storage_profile_name)
+            sp = etree.XML(etree.tostring(storage_profile, pretty_print=True))
+            sp_href = sp.attrib['href']
+            diskParms.append(E.StorageProfile(href=sp_href,
+                                              name=storage_profile_name))
         if disk is None:
```
rocknes: Move line 424-427 and 435-436 to the top of the function. Flow of the function should be
1. Get the existing disk object. If any error occurs here, we should bail as soon as possible.
2. Construct the new disk param
3. Make the rest call.

pyvcloud/vcd/vdc.py
sompa: changing this is as part of another CL.

pyvcloud/vcd/vdc.py
sompa: I will change this in another CL, since this does not break anything.

pyvcloud/vcd/vdc.py
```diff
@@ -415,6 +426,12 @@ def update_disk(self, name, size, new_name=None, description=None,
         else:
             disk = self.get_disk(name)
```
rocknes: I can see a bug in the implementation fo get_disk(name).

If there are two disks with the same name, we will return the first disk and never notify the user that there are more than 1 disk with the same name. We can address that in a new CL, but file a bug for the same.

sompa: Hey ! Get disk does take disk_id as a param. 
get_disk(self, name, disk_id=None):
The use case you are pointing is already taken care of and this was demoed too by Paco (Same name disks with different disk_id)

rocknes: I am pointing to the case where there are two disks with the same name, and user supplies only the name of the vm (no vm_id is provided). In that case we just iterate over the disks in the vdc and the first disk that matches the name filter is returned, without ever prompting to the user that there are two disks with the same name.

As a CLI user you are not expected to remember how many disks are there in your inventory every time you run a command. Our CLI should point out that info to the user without updating a non desired disk.

What Paco demonstrated was the ability to distinguish between two disks with the same name. But what is missing is that we are not alerting/stopping the user from modifying a non-desired vm, which happens to be namesake of the vm we actually want to modify.

PR #127, opened 2017-11-15
Commands for changing owner for vApp and independent disks

sahithi:
Commands for changing owner for vApp and independent disks

### DIFF
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -119,25 +119,28 @@ class Org(object):
         raise Exception('Catalog not found.')
 
     def get_user(self, user_name):
+        """
+        Retrieve user record from current Organization
+        :param user_name: user name of the record to be retrieved
+        :return: User record
+        """ # NOQA
         if self.resource is None:
             self.resource = self.client.get_resource(self.href)
         resource_type = 'adminUser' if self.is_admin else 'user'
-        org_filter = 'org==%s' % self.resource.get('href') if self.is_admin else None
-        result = []
-        q = self.client.get_typed_query(
+        org_filter = 'org==%s' % self.resource.get('href') \
+            if self.is_admin else None
+        query = self.client.get_typed_query(
             resource_type,
             query_result_format=QueryResultFormat.REFERENCES,
             equality_filter=('name', user_name),
             qfilter=org_filter
             )
-        records = list(q.execute())
+        records = list(query.execute())
         if len(records) == 0:
             raise Exception('user not found')
         elif len(records) > 1:
             raise Exception('multiple users found')
         return self.client.get_resource(records[0].get('href'))
-
-
 
     def share_catalog(self, name, share=True):
         catalog = self.get_catalog(name)
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -118,6 +118,27 @@ class Org(object):
                 return self.client.get_resource(link.href)
         raise Exception('Catalog not found.')
 
+    def get_user(self, user_name):
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+        resource_type = 'adminUser' if self.is_admin else 'user'
+        org_filter = 'org==%s' % self.resource.get('href') if self.is_admin else None
+        result = []
+        q = self.client.get_typed_query(
+            resource_type,
+            query_result_format=QueryResultFormat.REFERENCES,
+            equality_filter=('name', user_name),
+            qfilter=org_filter
+            )
+        records = list(q.execute())
+        if len(records) == 0:
+            raise Exception('user not found')
+        elif len(records) > 1:
+            raise Exception('multiple users found')
+        return self.client.get_resource(records[0].get('href'))
+
+
+
     def share_catalog(self, name, share=True):
         catalog = self.get_catalog(name)
         is_published = 'true' if share else 'false'
--- a/pyvcloud/vcd/vapp.py
+++ b/pyvcloud/vcd/vapp.py
@@ -124,6 +124,18 @@ class VApp(object):
             new_section,
             EntityType.LEASE_SETTINGS.value)
 
+    def change_owner(self, href):
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+        new_owner = self.resource.Owner
+        new_owner.User.set('href', href)
+        objectify.deannotate(new_owner)
+        etree.cleanup_namespaces(new_owner)
+        return self.client.put_resource(
+            self.resource.get('href') + '/owner/',
+            new_owner,
+            EntityType.OWNER.value)
+
     def power_off(self):
         if self.resource is None:
             self.resource = self.client.get_resource(self.href)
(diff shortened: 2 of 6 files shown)


pyvcloud/vcd/org.py
```diff
@@ -118,6 +118,27 @@ def get_catalog(self, name):
                 return self.client.get_resource(link.href)
         raise Exception('Catalog not found.')
 
+    def get_user(self, user_name):
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+        resource_type = 'adminUser' if self.is_admin else 'user'
+        org_filter = 'org==%s' % self.resource.get('href') if self.is_admin else None
```
anusuyar: Run flake8 and make sure formatting is correct

pip install -r test-requirements.txt
flake8 pyvcloud/vcd

pip install -r test-requirementes.txt
flake8 vcd_cli/*.py

pyvcloud/vcd/org.py
```diff
@@ -118,6 +118,27 @@ def get_catalog(self, name):
...
+        result = []
+        q = self.client.get_typed_query(
```
sompa: Can we use a little more descriptive variable name here than 'q' ?

pyvcloud/vcd/org.py
```diff
@@ -118,6 +118,27 @@ def get_catalog(self, name):
                 return self.client.get_resource(link.href)
         raise Exception('Catalog not found.')
 
+    def get_user(self, user_name):
```
sompa: Please add a doc string to every new functionality method you introduce in pyvcloud.

pyvcloud/vcd/vapp.py
```diff
@@ -124,6 +124,18 @@ def set_lease(self, deployment_lease=0, storage_lease=0):
             new_section,
             EntityType.LEASE_SETTINGS.value)
 
+    def change_owner(self, href):
```
sompa: Please add a doc string to every new functionality method you introduce in pyvcloud.

tests/vcd_disk.py
```diff
@@ -180,5 +180,16 @@ def test_010_get_storage_profile(self):
         print("Storage Profile: " + str(profile))
         assert profile.get('name') == self.config['vcd']['storage_profile']
 
+    def test_111_change_owner_disk(self):
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
+        vdc_resource = org.get_vdc(self.config['vcd']['vdc'])
+        user_resource = org.get_user(self.config['vcd']['new_disk_user'])
+        vdc = VDC(self.client, href=vdc_resource.get('href'))
+        disk = vdc.change_disk_owner(self.config['vcd']['disk'], user_resource.get('href'), self.config['vcd']['disk_id'])
```
sompa: Line too long. Can you run flake8 tests and verify that the line length is not > 80 ?

pyvcloud/vcd/vapp.py
```diff
@@ -124,6 +124,18 @@ def set_lease(self, deployment_lease=0, storage_lease=0):
...
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+        new_owner = self.resource.Owner
+        new_owner.User.set('href', href)
+        objectify.deannotate(new_owner)
```
rocknes: Doesn't objectify.deannotate and etree.cleanup_namespaces do the same job?

tests/vcd_disk.py
```diff
@@ -180,5 +180,16 @@ def test_010_get_storage_profile(self):
         print("Storage Profile: " + str(profile))
         assert profile.get('name') == self.config['vcd']['storage_profile']
 
+    def test_111_change_owner_disk(self):
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
```
rocknes: missing space around =

pyvcloud/vcd/vdc.py
```diff
@@ -509,6 +509,24 @@ def get_disk(self, name, disk_id=None):
...
+    def change_disk_owner(self, name, user_href, disk_id=None):
+
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+
+        if disk_id is not None:
+            disk = self.get_disk(None, disk_id)
+        else:
+            disk = self.get_disk(name)
+        new_owner = disk.Owner
```
rocknes: This works, but I think a better way would be to strip off the entire User section of the disk element and add back a new User element which has a single attribute viz. href.

pyvcloud/vcd/org.py
```diff
@@ -118,6 +118,27 @@ def get_catalog(self, name):
...
+        q = self.client.get_typed_query(
+            resource_type,
+            query_result_format=QueryResultFormat.REFERENCES,
+            equality_filter=('name', user_name),
+            qfilter=org_filter
+            )
+        records = list(q.execute())
+        if len(records) == 0:
+            raise Exception('user not found')
+        elif len(records) > 1:
```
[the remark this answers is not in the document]
sahithi: It can catch our own code bugs.
For example : When this method is invoked by Sys Admin to get an user with the same name in multiple organizations, it has to return unique result.

tests/vcd_disk.py
sahithi: Just followed the convention from other methods. No space.

pyvcloud/vcd/vdc.py
sahithi: User section has only two fields viz., name and href. I prefer to keep as-is this time as it doesn't make much difference. I will keep this in mind for future, thanks.

sahithi:
Addressed comments

pyvcloud/vcd/org.py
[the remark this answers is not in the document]
anusuyar: I don't think the change_ower command should fail when the same user is present two orgs(Even when invoked by sysadmin). I think we should restrict the user search to the 'org_in_use' even when invoked by sysadmin.

What if we had 3 org
system
org1 with userA1, userB1 with vapp1
org2 with userA2, userB2, userA1 with vaap2

We login as sysadmin
switch to org1

1. invoke change_owner(userA2, vaap1)
Will it now set userA2 as owner of vapp1 ? (userA2 is not a user in org1 and should fail)

2. invoke change_owner(userA1, vaap1)
   Should we fail this command since now get_user() will return 2 users and throw an exception? I don't think we should fail in this case. There could be common username between many orgs and this should not be the reason to fail the command.

I think we should have org_filter to the current 'org_in_use' for both sysadmin and nonadmin case.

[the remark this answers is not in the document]
sahithi: Anusuya, 

1. invoke change_owner(userA2, vapp1) - this will fail.
2. invoke change_owner(userA1, vapp1) - this will not fail and will correctly set the user with that of org-in-use.

org_filter is always set to org_in_use. Refer to vcd-cli/vapp/change_owner method.
_______
My point earlier was if ever in future, developer makes a mistake in passing the incorrect values from vcd-cli, the above check for more than one record will catch it.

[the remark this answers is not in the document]
sahithi: Sorry for not being clear in my earlier comment.

PR #129, opened 2017-11-17
Change Catalog Owner

sahithi:
I have no idea why my previous commits/merge are being shown here in this pull request.
Files changed relevant to catalog ownership:
1. org.py
2. vcd_catalog.py (tests)

### DIFF
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -25,7 +25,9 @@ from pyvcloud.vcd.client import get_links
 from pyvcloud.vcd.client import MissingRecordException
 from pyvcloud.vcd.client import QueryResultFormat
 from pyvcloud.vcd.client import RelationType
+from pyvcloud.vcd.utils import access_settings_to_dict
 from pyvcloud.vcd.utils import to_dict
+
 import shutil
 import tarfile
 import tempfile
@@ -141,7 +143,7 @@ class Org(object):
         elif len(records) > 1:
             raise Exception('multiple users found')
         return self.client.get_resource(records[0].get('href'))
-      
+
     def update_catalog(self, old_catalog_name, new_catalog_name, description):
         """
         Update the name and/or description of a catalog.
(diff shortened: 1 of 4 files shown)


pyvcloud/vcd/org.py
```diff
@@ -546,3 +554,25 @@ def get_roles_query(self, name_filter=None):
...
+        """
+        Change the ownership of Catalog to a given user
+        :param catalog_name: Catalog whose ownership needs to be changed
+        :param user_name: New Owner of the Catalog
+        :return: None
+        """ # NOQA
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+        catalog_resource = self.get_catalog_resource(catalog_name, True)
+        catalog_href = catalog_resource.get('href') + '/owner'
```
rocknes: owner link should be extracted off the catalog resource

pyvcloud/vcd/org.py
```diff
@@ -546,3 +554,25 @@ def get_roles_query(self, name_filter=None):
...
+    def change_catalog_owner(self, catalog_name, user_name):
+        """
+        Change the ownership of Catalog to a given user
+        :param catalog_name: Catalog whose ownership needs to be changed
+        :param user_name: New Owner of the Catalog
+        :return: None
+        """ # NOQA
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+        catalog_resource = self.get_catalog_resource(catalog_name, True)
```
rocknes: Sending is_admin_operation=True all the time without actually checking if the current logged in user has the rights to run the command or not is a bit iffy. It would be nice if we have a check upfront that restrict usage of these commands to only admins.

sahithi: I agree, I did think of this. Currently there is no method we can check if logged-in-user is an org admin. Current org.is_admin determines if it is sysadmin or not. We can implement is_org_admin once we start working on roles. I will add a task to refactor this code and few other pieces of code where I see this gap. Let me know if you have any other ideas.

pyvcloud/vcd/org.py
sahithi: Done

PR #132, opened 2017-11-22
Catalog change owner

sahithi:
I had to create a new pull request after resolving few issues. Sorry for losing reviews on previous pull request. I have addressed all of them though.

### DIFF
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -111,14 +111,21 @@ class Org(object):
         return result
 
     def get_catalog(self, name):
+        return self.get_catalog_resource(name, False)
+
+    def get_catalog_resource(self, name, is_admin_operation=False):
         org = self.client.get_resource(self.href)
         links = get_links(org,
                           rel=RelationType.DOWN,
                           media_type=EntityType.CATALOG.value)
         for link in links:
             if name == link.name:
-                return self.client.get_resource(link.href)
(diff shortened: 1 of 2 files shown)


pyvcloud/vcd/org.py
```diff
@@ -111,14 +111,21 @@ def list_catalogs(self):
...
 
     def get_catalog(self, name):
+        return self.get_catalog_resource(name, False)
+
+    def get_catalog_resource(self, name, is_admin_operation=False):
         org = self.client.get_resource(self.href)
         links = get_links(org,
                           rel=RelationType.DOWN,
                           media_type=EntityType.CATALOG.value)
         for link in links:
```
rocknes: We should probably extract this portion out and refactor it in a future commit as a filter service to get_links() call.

tests/vcd_catalog.py
```diff
@@ -22,5 +22,14 @@ def test_catalog_control_access_retrieval(self):
         control_access = org.get_catalog_access_control_settings(catalog.get('name'))
         assert control_access is not None
 
+    def test_change_catalog_owner(self):
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
```
rocknes: I was under the impression that org-admin and sys-admin can only change the owner of a catalog. If that is the case, should the Org object be created with is_admin=True?

PR #139, opened 2017-12-03
Refactoring is_admin field in org.py to client.py

rocknes:
Searched across all python files under \pyvcloud\pyvcloud\vcd and \pyvcloud\tests to make sure that all instances of org.is_admin has been replaced with the new logic. Additionally ran flake8 validation over the entire project to make sure that nothing is broken.

### DIFF
--- a/pyvcloud/vcd/client.py
+++ b/pyvcloud/vcd/client.py
@@ -25,7 +25,6 @@ import sys
 import time
 import urllib
 
-
 SIZE_1MB = 1024*1024
 
 
@@ -515,6 +514,8 @@ class Client(object):
 
         self.fsencoding = sys.getfilesystemencoding()
 
+        self._is_loggedin_user_sys_admin = False
+
     def _get_response_request_id(self, response):
         return response.headers[self._REQUEST_ID_HDR_NAME]
 
@@ -574,9 +575,18 @@ class Client(object):
         self._session.headers['x-vcloud-authorization'] = \
             response.headers['x-vcloud-authorization']
 
+        if session.get('org') == 'System':
+            self._is_loggedin_user_sys_admin = True
+        else:
+            self._is_loggedin_user_sys_admin = False
+
     def rehydrate(self, state):
         self._session = requests.Session()
         self._session.headers['x-vcloud-authorization'] = state.get('token')
+        if state.get('org') == 'System':
+            self._is_loggedin_user_sys_admin = True
+        else:
+            self._is_loggedin_user_sys_admin = False
         wkep = state.get('wkep')
         self._session_endpoints = {}
         for endpoint in _WellKnownEndpoint:
@@ -600,6 +610,12 @@ class Client(object):
                 Exception("Unknown login failure")
 
         session = objectify.fromstring(response.content)
+
+        if session.get('org') == 'System':
+            self._is_loggedin_user_sys_admin = True
+        else:
+            self._is_loggedin_user_sys_admin = False
+
         self._session_endpoints = _get_session_endpoints(session)
         self._session = new_session
         self._session.headers['x-vcloud-authorization'] = \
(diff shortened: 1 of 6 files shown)


pyvcloud/vcd/client.py
```diff
@@ -574,9 +575,18 @@ def set_credentials(self, creds):
         self._session.headers['x-vcloud-authorization'] = \
             response.headers['x-vcloud-authorization']
 
+        if session.get('org') == 'System':
+            self._is_loggedin_user_sys_admin = True
```
sahithi: self._is_loggedin_user_sys_admin = True if (session.get('org') == System) else False

pyvcloud/vcd/client.py
```diff
@@ -574,9 +575,18 @@ def set_credentials(self, creds):
...
+        else:
+            self._is_loggedin_user_sys_admin = False
+
     def rehydrate(self, state):
         self._session = requests.Session()
         self._session.headers['x-vcloud-authorization'] = state.get('token')
+        if state.get('org') == 'System':
+            self._is_loggedin_user_sys_admin = True
```
sahithi: Use ternary condition operator as mentioned above

pyvcloud/vcd/client.py
```diff
@@ -610,6 +626,9 @@ def logout(self):
         uri = self._uri + '/session'
         return self._do_request('DELETE', uri)
 
+    def is_loogedin_user_sys_admin(self):
```
sahithi: spell check - loogedin

pyvcloud/vcd/client.py
```diff
@@ -600,6 +610,12 @@ def rehydrate_from_token(self, token):
                 Exception("Unknown login failure")
 
         session = objectify.fromstring(response.content)
+
+        if session.get('org') == 'System':
+            self._is_loggedin_user_sys_admin = True
```
sahithi: ternary condition

pyvcloud/vcd/client.py
```diff
@@ -515,6 +514,8 @@ def __init__(self,
 
         self.fsencoding = sys.getfilesystemencoding()
 
+        self._is_loggedin_user_sys_admin = False
```
sahithi: variable seems too long. As we can expect this to be used extensively, can we shorten the name? how about is_sysadmin or anything better?

pyvcloud/vcd/client.py
sahithi: As this method going to be used extensively at several places, please shorten the name. is_sysadmin or better?

pyvcloud/vcd/client.py
anusuyaceg: we don't need to set the variable to False, as it is already assigned as False as default. The If statement is more readable in this case.

pyvcloud/vcd/client.py
anusuyaceg: +1

anusuyaceg: is_user_sysadmin? I don't think, just because the method will be used in many places, it should be shortened. 
 If we are shortening it, we should add docstring to describe the method.

pyvcloud/vcd/client.py
rocknes: I don't see a reason why we should stick to ternary operator. Infact the styling of python ternary operator has been a topic of ragging debates where people have raised it's syntax in question. In short I feel that the simple if statement is clearer to read and doesn't comprise in any way.

ref : https://stackoverflow.com/questions/394809/does-python-have-a-ternary-conditional-operator

pyvcloud/vcd/client.py
rocknes: oops, my bad

rocknes: I agree that the name of the variable is a mouthful.
What we really want to ascertain is that the current context of the logged in user is that of sys_admin.

client.is_in_sysadmin_context() makes more sense?

pyvcloud/vcd/client.py
sahithi: okay.

pyvcloud/vcd/client.py
sahithi: Anusuya, it is better to have short method names in general. Given that this method is going to be used extensively, imho, it is okay to give it a second thought.

anusuyaceg: @sahithi okay.

PR #140, opened 2017-12-04
List rights for a given role.

sahithi:
Command sample output:
> vcd role list-rights 'vApp User'
name
---------------------------------
vApp Template: Checkout
UI Plugins: View
vApp: Snapshot Operations
vApp: Copy
vApp: Manage VM Password Settings
vApp Template / Media: View
vApp: Edit VM Properties
vApp: View VM metrics
vApp: Delete
vApp: Use Console
Disk: View Properties
vApp: Power Operations
vApp: Sharing
vApp: Edit Properties
vApp: Edit VM Network

### DIFF
--- a/pyvcloud/vcd/role.py
+++ b/pyvcloud/vcd/role.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from vcd_cli.utils import stdout
 
 class Role(object):
     def __init__(self, client, href=None, resource=None):
@@ -31,7 +32,7 @@ class Role(object):
 
     def list_rights(self):
         """
-        List rights of the role
+        List rights associated with the role
         :return: list of names of rights for a given role.
         """  # NOQA
         if self.resource is None:
@@ -41,5 +42,4 @@ class Role(object):
                 hasattr(self.resource.RightReferences, 'RightReference'):
             for right in self.resource.RightReferences.RightReference:
                 rights.append({'name': right.get('name')})
-            return rights
-        raise Exception("No rights found for the role \'%s\'" % self.name)
+        return rights
--- a/pyvcloud/vcd/role.py
+++ b/pyvcloud/vcd/role.py
@@ -1,0 +1,41 @@
+# VMware vCloud Director Python SDK
+# Copyright (c) 2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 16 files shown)


pyvcloud/vcd/role.py
```diff
@@ -0,0 +1,45 @@
...
+        """  # NOQA
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+        rights = []
+        if hasattr(self.resource, 'RightReferences') and \
+                hasattr(self.resource.RightReferences, 'RightReference'):
+            for right in self.resource.RightReferences.RightReference:
+                rights.append({'name': right.get('name')})
+            return rights
+        raise Exception("No rights found for the role \'%s\'" % self.name)
```
anusuyar: Should we throw an exception when there are no rights in the role?(is it possible to have role with no rights?)  I think it's better to return the rights even if it is empty.

pyvcloud/vcd/role.py
```diff
@@ -0,0 +1,45 @@
...
+        self.client = client
+        self.href = href
+        self.resource = resource
+        if resource is not None:
+            self.href = resource.get('href')
+            self.name = resource.get('name')
+
+    def list_rights(self):
+        """
+        List rights of the role
```
rocknes: nit : List rights associated with the role

pyvcloud/vcd/role.py
```diff
@@ -0,0 +1,45 @@
...
+        :return: list of names of rights for a given role.
```
rocknes: Shouldn't we also send back the id of the right?

pyvcloud/vcd/role.py
rocknes: ^ +1.

pyvcloud/vcd/role.py
sahithi: done

pyvcloud/vcd/role.py
sahithi: Valid point. Modified the code.

pyvcloud/vcd/role.py
sahithi: User can retrieve ids using search command. 
I quickly checked all other commands with list sub-command, none of them is listing ids of any resources.

sahithi:
Addressed comments

pyvcloud/vcd/role.py
sahithi: done

pyvcloud/vcd/role.py
sahithi: None of the existing list subcommands' (org, vapp etc) lists ids.
Ids can always be retrieved through search command.

pyvcloud/vcd/role.py
sahithi: Fixed it.

PR #145, opened 2017-12-07
Create a role in the specified Org or current org in use

sahithi:
Inspected code as per pep8 standards.

### DIFF
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -76,6 +76,33 @@ class Org(object):
             EntityType.ADMIN_CATALOG.value,
             catalog)
 
+    def create_role(self, role_name, description, rights):
+        """
+        Creates a role in the organization
+        :param role_name (str): name of the role to be created
+        :param description (str): description of the role
+        :param rights (list): names of zero or more rights to be associated with the role
+        :return: RoleType just created
+        """ #  NOQA
+        org_admin_resource = self.client.get_resource(self.href_admin)
+        role = E.Role(
+            E.Description(description),
+            E.RightReferences(),
+            name=role_name
+        )
+        for right in rights:
+            right_record = self.get_right(right)
+            role.RightReferences.append(
+                E.RightReference(name=right_record.get('name'),
+                                 href=right_record.get('href'),
+                                 type=EntityType.RIGHT.value))
+        return self.client.post_linked_resource(
+            org_admin_resource,
+            RelationType.ADD,
+            EntityType.ROLE.value,
+            role)
+
+
     def delete_catalog(self, name):
         org = self.client.get_resource(self.href)
         links = get_links(org,
@@ -563,6 +590,39 @@ class Org(object):
             query_result_format=QueryResultFormat.RECORDS,
             equality_filter=name_filter,
             qfilter=org_filter)
+        return query, resource_type
+
+    def get_right(self, right_name):
+        """
+        Retrieves corresponding record of the specified right.
+        :param right_name: (str): The name of the right record to be retrieved
+        :return: (QueryResultRightRecordType): Right query result in records
+                 format
+        """  # NOQA
+        try:
+            rights_query = self.get_rights_query(('name', right_name))[0]
+            return rights_query.find_unique()
+        except MissingRecordException:
+            raise Exception('Right \'%s\' does not exist.' % right_name)
+
+    def get_rights_query(self, name_filter=None):
+        """
+        Get the typed query for the rights in the current Org
+        :param name_filter: (tuple): (name ,'right name') Filter the rights by
+                             'right name'
+        :return: (tuple of (_TypedQuery, str))
+                  _TypedQuery object represents the query for the rights in
+                  the current Org
+                  str represents the resource type of the query object
+        """  # NOQA
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+
+        resource_type = 'right'
+        query = self.client.get_typed_query(
+            resource_type,
+            query_result_format=QueryResultFormat.RECORDS,
+            equality_filter=name_filter)
         return query, resource_type
 
     def get_catalog_access_control_settings(self, catalog_name):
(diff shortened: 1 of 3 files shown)


pyvcloud/vcd/org.py
```diff
@@ -77,6 +77,32 @@ def create_catalog(self, name, description):
             EntityType.ADMIN_CATALOG.value,
             catalog)
 
+    def create_role(self, role_name, description, rights):
+        """
+        Creates a role in the organization
+        :param role_name: name of the role to be created
```
anusuyar: Let's also specify the type of each argument. 
like 
:param role_name:  (str): name of the role to be created

pyvcloud/vcd/org.py
```diff
@@ -582,6 +608,39 @@ def get_roles_query(self, name_filter=None):
...
+        """  # NOQA
+        if self.resource is None:
+            self.resource = self.client.get_resource(self.href)
+
+        resource_type = 'right'
+        query = self.client.get_typed_query(
+            resource_type,
+            query_result_format=QueryResultFormat.RECORDS,
+            equality_filter=name_filter)
+        return query, resource_type
```
anusuyar: Why are we passing resource_type here? I dont see it being used in the calling function.

tests/vcd_role.py
```diff
@@ -26,5 +26,13 @@ def test_get_rights(self):
         rights = role.list_rights()
         assert len(rights) > 0
 
+    def test_create_role(self):
```
anusuyar: We can add copyrights header for test files too.

pyvcloud/vcd/org.py
sahithi: It is going to be used in another change-list underway (list rights in the org).

pyvcloud/vcd/org.py
```diff
@@ -77,6 +77,32 @@ def create_catalog(self, name, description):
...
+        :param rights: names of zero or more rights to be associated with the role
+        :return: RoleType just created
+        """  # NOQA
+        org_admin_resource = self.client.get_resource(self.href_admin)
+        role = E.Role(
+            E.Description(description),
+            E.RightReferences(),
+            name=role_name
+        )
+        for right in rights:
```
rocknes: Before iterating, I think we should make sure that rights is a list. Also a Null check should be made.

tests/vcd_role.py
sahithi: fixed.

pyvcloud/vcd/org.py
```diff
@@ -582,6 +608,39 @@ def get_roles_query(self, name_filter=None):
...
+        :return: (QueryResultRightRecordType): Right query result in records
+                 format
+        """  # NOQA
+        try:
+            rights_query = self.get_rights_query(('name', right_name))[0]
+            return rights_query.find_unique()
+        except MissingRecordException:
+            raise Exception('Right \'%s\' does not exist.' % right_name)
+
+    def get_rights_query(self, name_filter=None):
```
rocknes: This looks out of place. Query related stuff should be in client.py

pyvcloud/vcd/org.py
rocknes: I don't think we should be passing back resource_type unless we make get_rights_query a generic function. The name of the function is a dead giveaway that we are dealing with 'rights'.

tests/vcd_role.py
```diff
@@ -26,5 +41,13 @@ def test_get_rights(self):
...
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
+        role_name = self.config['vcd']['role_name']
+        org.create_role(role_name, 'test role', {'vApp: Copy'})
```
pacogomez: If the method is expecting a tuple of strings, this should be `, ('vApp: Copy'))`, not a dict

pyvcloud/vcd/org.py
pacogomez: We have `list_*`  methods outside client.py. Maybe this should renamed to `list_rights`?

pyvcloud/vcd/org.py
pacogomez: This should only return one value, the list of rights, see `list_catalogs` as an example

tests/vcd_role.py
sahithi: my mistake, fixed.

pyvcloud/vcd/org.py
sahithi: I changed:  for right in rights -> for right in tuple(rights). This doesn't require None check.

pyvcloud/vcd/org.py
```diff
@@ -582,6 +610,39 @@ def get_roles_query(self, name_filter=None):
...
+        """  # NOQA
+        right_record = self.list_rights(('name', right_name))
+        if len(right_record) < 1:
+            raise Exception('Right \'%s\' does not exist.' % right_name)
+        return right_record[0]
+
+    def list_rights(self, name_filter=None):
+        """
+        Get the typed query for the rights in the current Org
+        :param name_filter: (tuple): (name ,'right name') Filter the rights by
```
rocknes: I don't think we need a tuple here, rather a support for multiple filters (by name) which can be logically or-ed will be more valuable. 

fyi : Rights are unique in context of an org, so this list will always contain 1 or 0.

tests/vcd_role.py
```diff
@@ -26,5 +41,13 @@ def test_get_rights(self):
         rights = role.list_rights()
         assert len(rights) > 0
 
+    def test_create_role(self):
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
+        role_name = self.config['vcd']['role_name']
+        org.create_role(role_name, 'test description', ('Disk: View Properties',))
+        role_record = org.get_role(role_name)
+        assert self.config['vcd']['role_name'] == role_record.get('name')
```
rocknes: Please make sure you clean up the role after the tests are done.

sahithi: Yes, I will make a not to modify this test once I implement the command to delete a role.

PR #150, opened 2017-12-12
List all rights in an organization

sahithi:


anusuyar:
Renamed which file? I am seeing only the added unit tests.

PR #167, opened 2018-01-05
Add or Remove rights from a given role

sahithi:
Add or Remove rights from a given role
Unittests

Example:
To begin with, Role 'vApp User' has zero rights.

    >vcd role add-right 'vApp User' 'vApp: Copy' 'vApp: Edit VM CPU'
    List of rights in the role 'vApp User':
    name
    ========
    vApp: Copy
    vApp: Edit VM CPU

    >vcd role remove-right 'vApp User' 'vApp: Copy'
    List of rights in the role 'vApp User':
    name
    =======
    vApp: Edit VM CPU

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/167)


sompa:
*[tests/vcd_role_right.py, line 103 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L26j2dm-vuuje0uWHo1:-L26j2dm-vuuje0uWHo2:begblrc) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L103)):*
> ```Python
>     def test_10_remove_rights_from_role(self):
>         org_in_use = self.config['vcd']['org_in_use']
>         org = Org(self.client, href = self.client.get_org_by_name(org_in_use).get('href'))
> ```

flake8 : line length > 80.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

sompa:
*[pyvcloud/vcd/role.py, line 99 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L26jV33QsYcr7DR2QUz:-L26jV33QsYcr7DR2QV-:b-ysns7h) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L99)):*
> ```Python
>         :return: A :class:`lxml.objectify.StringElement` object representing
>                 the updated role resource.
>         """
> ```

Requires a new line here or add #NOQA to avoid flake8.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

sompa:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

pacogomez:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 99 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L26jV33QsYcr7DR2QUz:-L27-T7pE9ZomY4LUG7R:b-o5mwrj) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L99)):*
<details><summary><i>Previously, sompa (malakars) wrote…</i></summary><blockquote>

Requires a new line here or add #NOQA to avoid flake8.
</blockquote></details>

It is preferred to format the docstring as recommended and not use #NOQA

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

pacogomez:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 83 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L276DLE0hTb3QzDG6YD:-L276DLE0hTb3QzDG6YE:b-u3tqpl) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L83)):*
> ```Python
>             self.resource = self.client.get_resource(self.href)
>         updated_resource = self.resource
>         for right_record in list(right_records):
> ```

is there a need to make it a list? it is supposed to be a list

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

anusuyar:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 102 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27YGqir5eDGBc7JUUN:-L27YGqir5eDGBc7JUUO:bpz5ize) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L102)):*
> ```Python
>         if self.resource is None:
>             self.resource = self.client.get_resource(self.href)
>         rights = list(rights)
> ```

This might be unnecessary if rights is already a list.

---

*[pyvcloud/vcd/role.py, line 103 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27_Q9EEpBehSKFNc68:-L27_Q9EEpBehSKFNc69:belo2sh) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L103)):*
> ```Python
>             self.resource = self.client.get_resource(self.href)
>         rights = list(rights)
>         updated_resource = self.resource
> ```

can we name this more sepecific> Like updated_role_resource?

---

*[pyvcloud/vcd/role.py, line 116 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27ZtTGxezSBR_EoT5r:-L27ZtTGxezSBR_EoT5s:bwdgodp) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L116)):*
> ```Python
>                     break
>         if len(rights) > 0:
>             print('Skipping removal of rights [', ', '.join(rights), ']\
> ```

Should we allow the remove operation to fail if at least one right does not exist? or just skip it? What do others think about this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

rocknes:
In pyvcloud pull requests, instead of vcd-cli command outputs,  we should demonstrate that relevant unit tests have passed.

---

Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 75 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G67iyKa4TG5N6aQHx:-L2G67iyKa4TG5N6aQHy:b-mm47qu) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L75)):*
> ```Python
>         """Adds list of rights to a given role
> 
>         :param: right_records: (list): List of right records
> ```

We should add a little detail on the structure of the right_record object we are expecting viz. presence of the fields, 'href' and 'name'.

It's likely that consumer has access to href and name of a right but not the entire record object obtained via query, we should clarify that it's ok to substitute one for another.

---

*[pyvcloud/vcd/role.py, line 89 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G-f7f1lPsXWraUtsB:-L2G-f7f1lPsXWraUtsC:b-bzleh5) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L89)):*
> ```Python
>                     href=right_record.get('href'),
>                     type=EntityType.RIGHT.value))
>         return self.client.put_resource(
> ```

It's better if we assign self.resource the returned value and then return self.resource. It can potentially save a resource reload by the consumer.

Also, I think we should start the habit of passing param names along with the param values (I know I am guilty of not doing it always too :) ). Python lacks type checking, relying on arg ordering can lead to nasty surprises down the line, if we ever modify the arg list of these multi-arg functions.

---

*[pyvcloud/vcd/role.py, line 106 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G1XVc2rk7e5xc6zrs:-L2G1XVc2rk7e5xc6zrt:bk1wdz0) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L106)):*
> ```Python
>         if hasattr(self.resource, 'RightReferences') and \
>                 hasattr(self.resource.RightReferences, 'RightReference'):
>             rightReferenceList = self.resource.RightReferences.RightReference
> ```

It's better to ditch the intermediate variables, if they are not used more than once.

---

*[pyvcloud/vcd/role.py, line 111 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G22xH62HbKQuBc5G4:-L2G22xH62HbKQuBc5G5:bptomrp) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L111)):*
> ```Python
>                     if rightReference.get('name') == right:
>                         updated_resource.RightReferences.remove(rightReference)
>                         rights.remove(right)
> ```

It's generally a bad idea to change the data structure on which an iterator is currently running.

---

*[pyvcloud/vcd/role.py, line 116 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27ZtTGxezSBR_EoT5r:-L2G2iXCNHmNZ-od_XNL:b2e3k0n) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L116)):*
<details><summary><i>Previously, anusuyar wrote…</i></summary><blockquote>

Should we allow the remove operation to fail if at least one right does not exist? or just skip it? What do others think about this?
</blockquote></details>

If rights, specified by user for removal, are missing from the role, I think we should just highlight that fact and proceed with  whatever rights we can remove.

Message can be a bit crisper,
Warning : Rights {list of rights} are not associated with role <role name>.

---

*[pyvcloud/vcd/role.py, line 118 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G4AXIMm6cf5T3juKo:-L2G4AXIMm6cf5T3juKp:byfuv0e) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L118)):*
> ```Python
>             print('Skipping removal of rights [', ', '.join(rights), ']\
>              as they do not exist in the role')
>         return self.client.put_resource(
> ```

Same suggestion as above, if we are getting the updated resource for free, we should update the resource before sending it back.

---

*[tests/vcd_role_right.py, line 82 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G57FJlQbpzTWny13P:-L2G4zQlG22d7GP5BK2b:bkddk1s) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L82)):*
> ```Python
>     def test_09_add_rights_To_Role(self):
>         org_in_use = self.config['vcd']['org_in_use']
>         org = Org(self.client, href=self.client.get_org_by_name(org_in_use).get('href'))
> ```

probably line length > 80

---

*[tests/vcd_role_right.py, line 96 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G7lGnywhCKCdhZe3_:-L2G7lGnywhCKCdhZe3a:b-iqwd5z) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L96)):*
> ```Python
>                 hasattr(updated_role_resource.RightReferences, 'RightReference'):
>             for right in updated_role_resource.RightReferences.RightReference:
>                 if right.get('name') == right_name:
> ```

This test will be valid only if we know that the right was not associated with the role to begin with.
We should run the same for loop on role.resource object before calling add_rights, and put an assert accordingly.

---

*[tests/vcd_role_right.py, line 111 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G71Etl2EXiS2CXY__:-L2G71Etl2EXiS2CXY_a:be4v3ms) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L111)):*
> ```Python
> 
>         updated_role_resource = role.remove_rights([right_name])
>         success = True
> ```

This test will be valid only if we know that the right was associated with the role to begin with.

We should run the same for loop on role.resource object before calling remove_rights, success = True if we find the right else assert and break the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-:-L2Fzrc6u6nlBcruilMm:b-67ki8n)*

sahithi:
Review status: 0 of 2 files reviewed at latest revision, 14 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 75 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G67iyKa4TG5N6aQHx:-L2MVLR9Bsz1PqA0Wh8y:b-896fix) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L75)):*

We should add a little detail on the structure of the right_record object we are expecting viz. presence of the fields, 'href' and 'name'.

It's likely that consumer has access to href and name of a right but not the entire record object obtained via query, we should clarify that it's ok to substitute one for another.
</blockquote></details>

Done.

---

*[pyvcloud/vcd/role.py, line 83 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L276DLE0hTb3QzDG6YD:-L2MVXv3MHM3XGnBm8VK:b-fwrv3k) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L83)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

is there a need to make it a list? it is supposed to be a list
</blockquote></details>

I re-wrapped it in the list as a safe side. For example if user passes one right name instead of a list it protects its from failing.
Anyway, I removed it as you all think it is redundant.

---

*[pyvcloud/vcd/role.py, line 89 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G-f7f1lPsXWraUtsB:-L2MXBSnP0RKYy8PaSi2:b-896fix) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L89)):*

It's better if we assign self.resource the returned value and then return self.resource. It can potentially save a resource reload by the consumer.

Also, I think we should start the habit of passing param names along with the param values (I know I am guilty of not doing it always too :) ). Python lacks type checking, relying on arg ordering can lead to nasty surprises down the line, if we ever modify the arg list of these multi-arg functions.
</blockquote></details>

Done.

---

*[pyvcloud/vcd/role.py, line 102 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27YGqir5eDGBc7JUUN:-L2MXL2-p3H7TNjnJlxf:b-riopa5) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L102)):*
<details><summary><i>Previously, anusuyar wrote…</i></summary><blockquote>

This might be unnecessary if rights is already a list.
</blockquote></details>

It is passed in as a tuple and tuple is immutable. There is a need for removal of rights in the for loop. So, I converted it to list.

---

*[pyvcloud/vcd/role.py, line 106 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G1XVc2rk7e5xc6zrs:-L2MXiNvz4EZAJYNGNlo:b-3t5go4) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L106)):*

It's better to ditch the intermediate variables, if they are not used more than once.
</blockquote></details>

I agree in general, but for this particular case, I added it for readability sake and would prefer to keep it.

---

*[pyvcloud/vcd/role.py, line 111 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G22xH62HbKQuBc5G4:-L2MdvcUTlADM3PvZlWD:b-896fix) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L111)):*

It's generally a bad idea to change the data structure on which an iterator is currently running.
</blockquote></details>

Done.

---

*[pyvcloud/vcd/role.py, line 116 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27ZtTGxezSBR_EoT5r:-L2MZ7jKOdpSsWTVFeqt:bs7wumb) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L116)):*

If rights, specified by user for removal, are missing from the role, I think we should just highlight that fact and proceed with  whatever rights we can remove.

Message can be a bit crisper,
Warning : Rights {list of rights} are not associated with role <role name>.
</blockquote></details>

Done.

---

*[pyvcloud/vcd/role.py, line 118 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G4AXIMm6cf5T3juKo:-L2MZBMj5SW7KvRm37ac:b-896fix) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L118)):*

Same suggestion as above, if we are getting the updated resource for free, we should update the resource before sending it back.
</blockquote></details>

Done.

---

*[tests/vcd_role_right.py, line 82 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G57FJlQbpzTWny13P:-L2MdzWbnkV-uXIdPHvo:bszaway) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L82)):*

probably line length > 80
</blockquote></details>

No warnings from flake8

---

*[tests/vcd_role_right.py, line 96 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G7lGnywhCKCdhZe3_:-L2Meg1MwGa-C1fDrst9:b-fevcrt) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L96)):*

This test will be valid only if we know that the right was not associated with the role to begin with.
We should run the same for loop on role.resource object before calling add_rights, and put an assert accordingly.
</blockquote></details>

I think this is overkill for an unittest. All our unittests run with assumptions for all CRUD operations. 

---

*[tests/vcd_role_right.py, line 111 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G71Etl2EXiS2CXY__:-L2MfABOquo0TEHghGPd:bt5kaya) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L111)):*

This test will be valid only if we know that the right was associated with the role to begin with.

We should run the same for loop on role.resource object before calling remove_rights, success = True if we find the right else assert and break the test.
</blockquote></details>

test_10_remove_rights_from_role() is placed such that it always runs after test_09_add_rights_To_Role(). So, it will only remove the right that has been added in the previous test. I noticed the same procedure in several existing unittests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

anusuyar:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

sahithi:
Review status: 1 of 2 files reviewed at latest revision, 13 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 103 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27_Q9EEpBehSKFNc68:-L2NDKl9LLkadktuAju_:b5rd511) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L103)):*
<details><summary><i>Previously, anusuyar wrote…</i></summary><blockquote>

can we name this more sepecific> Like updated_role_resource?
</blockquote></details>

Removed it altogether

---

*[tests/vcd_role_right.py, line 103 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L26j2dm-vuuje0uWHo1:-L2NDbA8Q3xwMajqPDy0:bszaway) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L103)):*
<details><summary><i>Previously, sompa (malakars) wrote…</i></summary><blockquote>

flake8 : line length > 80.
</blockquote></details>

No warnings from flake8

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

anusuyar:
:lgtm:

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-:-L2NGTjcp8EbZbL2mY-j:bnfp4nl)*

rocknes:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 102 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27YGqir5eDGBc7JUUN:-L2NimS_IRr1OxG5WsYG:b-inmzt5) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L102)):*
<details><summary><i>Previously, sahithi (Sahithi Ayloo) wrote…</i></summary><blockquote>

It is passed in as a tuple and tuple is immutable. There is a need for removal of rights in the for loop. So, I converted it to list.
</blockquote></details>

We have wrapped rights twice in a list. Once at
rights = list(rights)
and once again at for right in list(rights)

We can just create a separate variable for list(rights) and operate (call remove()) on it while iterating on the original tuple. That should cut down our wrapping to just 1 list.

---

*[tests/vcd_role_right.py, line 96 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2G7lGnywhCKCdhZe3_:-L2NjNI5dgq2MA_zAoQj:b-3tfvb2) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/tests/vcd_role_right.py#L96)):*
<details><summary><i>Previously, sahithi (Sahithi Ayloo) wrote…</i></summary><blockquote>

I think this is overkill for an unittest. All our unittests run with assumptions for all CRUD operations. 
</blockquote></details>

From what I understand about the currently implemented unit tests, we are making rest calls to vCD and from the http status value of the response, we are deciding the fate of the unit test. 

Since this test went a step ahead and checked for the presence of the newly added right in the updates xml representation, I pointed out what the complete test case should look like. Personally I feel, if we are making the effort to check things after the ReST call is complete, we should also check things before the call is made, to setup the control state of the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-:-L2NjJMWHhdXKckXQt4s:bnfp4nl)*

pacogomez:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 102 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27YGqir5eDGBc7JUUN:-L2OUVeDcxBuBV6Fu37J:b6vi7k3) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L102)):*

We have wrapped rights twice in a list. Once at
rights = list(rights)
and once again at for right in list(rights)

We can just create a separate variable for list(rights) and operate (call remove()) on it while iterating on the original tuple. That should cut down our wrapping to just 1 list.
</blockquote></details>

if there is a need to select a subset of the provided list it is better to use a separate variable with the selected items

---

*[pyvcloud/vcd/role.py, line 86 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OSVQ8lyF50qNRiYj4:-L2OSVQ8lyF50qNRiYj5:b-a9i5ja) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L86)):*
> ```Python
>             self.resource = self.client.get_resource(self.href)
>         for right in rights:
>             try:
> ```

- if the right doesn't exist or get_right raises an exception, what will it happen to the `right_record` variable? the try/except block swallows the exception and will make the lines below inaccurate or cause another error like right_record is not defined or None or with the previous value if any...
- this method should not print to stdout
- we probably should just remove the try/except

---

*[pyvcloud/vcd/role.py, line 122 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OUF3aN9JER8iboLpC:-L2OUF3aN9JER8iboLpD:b-cwof0z) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L122)):*
> ```Python
>                     break
>         if len(rights) > 0:
>             print("Skipping removal of rights [",
> ```

we shouldn't assume pyvcloud methods will be used only from the command line, so use of `print` in pyvcloud methods should be avoided. If a method needs to return any information to the caller, it should do it in the return method, or through an exception but not via `print`. Here it is probably fine to just return the result of `put_resource`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

sahithi:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 102 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27YGqir5eDGBc7JUUN:-L2RHZEi3xjLcIAHjA8q:bqcfnej) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L102)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

if there is a need to select a subset of the provided list it is better to use a separate variable with the selected items
</blockquote></details>

Aritra, Two list conversion serve different purposes. First one converts tuple into list so that we can remove elements from it. Second one is to address your comment to prevent removing elements from the list we are iterating.

To avoid confusion, I have removed the first safe check.

---

*[pyvcloud/vcd/role.py, line 86 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OSVQ8lyF50qNRiYj4:-L2RGqtangQ5azW-O9Go:b-771u3) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L86)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

- if the right doesn't exist or get_right raises an exception, what will it happen to the `right_record` variable? the try/except block swallows the exception and will make the lines below inaccurate or cause another error like right_record is not defined or None or with the previous value if any...
- this method should not print to stdout
- we probably should just remove the try/except

</blockquote></details>

Thanks for catching the error Paco. I missed adding "continue" statement. My test luckily passed becauseinvalid right was the second element in the list.

If we remove the try/except block entire operation would fail if there was one invalid entry.

---

*[pyvcloud/vcd/role.py, line 122 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OUF3aN9JER8iboLpC:-L2RIUfdXHQ01KJzryab:b1uty6b) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L122)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

we shouldn't assume pyvcloud methods will be used only from the command line, so use of `print` in pyvcloud methods should be avoided. If a method needs to return any information to the caller, it should do it in the return method, or through an exception but not via `print`. Here it is probably fine to just return the result of `put_resource`
</blockquote></details>

Any suggestions or existing standard if there is a need to communicate something to user.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

pacogomez:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 102 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27YGqir5eDGBc7JUUN:-L2R_dA6AsVSVj7Z8Gvm:bhiejle) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L102)):*
<details><summary><i>Previously, sahithi (Sahithi Ayloo) wrote…</i></summary><blockquote>

Aritra, Two list conversion serve different purposes. First one converts tuple into list so that we can remove elements from it. Second one is to address your comment to prevent removing elements from the list we are iterating.

To avoid confusion, I have removed the first safe check.
</blockquote></details>

see my comment above, I'm not sure we should use `self.resource`. Can you think about pros and cons of using `self.resource` instead of a temporary variable?

---

*[pyvcloud/vcd/role.py, line 86 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OSVQ8lyF50qNRiYj4:-L2RYJHmrha68LwStDis:bqjptjo) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L86)):*
<details><summary><i>Previously, sahithi (Sahithi Ayloo) wrote…</i></summary><blockquote>

Thanks for catching the error Paco. I missed adding "continue" statement. My test luckily passed becauseinvalid right was the second element in the list.

If we remove the try/except block entire operation would fail if there was one invalid entry.
</blockquote></details>

instead of populating `self.resource`, I think we should use a temporary variable

---

*[pyvcloud/vcd/role.py, line 122 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OUF3aN9JER8iboLpC:-L2RaQsbUQpkEZvS-ayW:b-tyavw0) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L122)):*
<details><summary><i>Previously, sahithi (Sahithi Ayloo) wrote…</i></summary><blockquote>

Any suggestions or existing standard if there is a need to communicate something to user.
</blockquote></details>

three options: 1. code can raise an exception if the right is not in the existing list 2. just continue processing without any processing, 3. return a tuple, where the second item is an error or warning message. I will prefer 1.

---

*[pyvcloud/vcd/role.py, line 91 at r4](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2RYeSEdSR7YtDK3Vy_:-L2RYeSEdSR7YtDK3Vya:bocu0e3) ([raw file](https://github.com/vmware/pyvcloud/blob/0f55221da3b3d55a4d26e27a045f2d6ba28e8146/pyvcloud/vcd/role.py#L91)):*
> ```Python
>                     href=right_record.get('href'),
>                     type=EntityType.RIGHT.value))
>         self.resource = self.client.put_resource(
> ```

while I understand the value of populating `self.resource`, if vCD replies with an error, the local cache might have inaccurate information anyways. It would be easier just to return the result of `put_resource` and not attempt to update `self.resource` by default

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

sahithi:
Review status: 0 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[pyvcloud/vcd/role.py, line 99 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L26jV33QsYcr7DR2QUz:-L2S7pH3v58aJ9-35s-4:b-896fix) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L99)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

It is preferred to format the docstring as recommended and not use #NOQA
</blockquote></details>

Done.

---

*[pyvcloud/vcd/role.py, line 102 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L27YGqir5eDGBc7JUUN:-L2RkH1lcnL44MMrEDBR:b-qwt713) ([raw file](https://github.com/vmware/pyvcloud/blob/d7cb4067669b3ad2a26f25dd57326bd4ae0cb868/pyvcloud/vcd/role.py#L102)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

see my comment above, I'm not sure we should use `self.resource`. Can you think about pros and cons of using `self.resource` instead of a temporary variable?
</blockquote></details>

I used temporary variable to do the same in my previous version of review. But I noticed that there is no difference at all as the variable is just a reference and any change to the data structure will reflect in both the references. Hence, I removed the temp variable.

---

*[pyvcloud/vcd/role.py, line 86 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OSVQ8lyF50qNRiYj4:-L2Rj9I4dU77YvXibguZ:b7d9tmg) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L86)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

instead of populating `self.resource`, I think we should use a temporary variable
</blockquote></details>

Yes, I used temporary variable to do the same in my previous version of review. But I noticed that there is no difference at all as the variable is just a reference and any change to the data structure will reflect in both the references. Hence, I removed the temp variable.

---

*[pyvcloud/vcd/role.py, line 122 at r3](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2OUF3aN9JER8iboLpC:-L2RizB882Bx9T5vargi:b-896fix) ([raw file](https://github.com/vmware/pyvcloud/blob/a5305c07d00a8564d882c41fa7a685882176684c/pyvcloud/vcd/role.py#L122)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

three options: 1. code can raise an exception if the right is not in the existing list 2. just continue processing without any processing, 3. return a tuple, where the second item is an error or warning message. I will prefer 1.
</blockquote></details>

Done.

---

*[pyvcloud/vcd/role.py, line 91 at r4](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-L2RYeSEdSR7YtDK3Vy_:-L2RjZVAwdRr60tRQHG8:b-onkmxv) ([raw file](https://github.com/vmware/pyvcloud/blob/0f55221da3b3d55a4d26e27a045f2d6ba28e8146/pyvcloud/vcd/role.py#L91)):*
<details><summary><i>Previously, pacogomez (Paco Gómez) wrote…</i></summary><blockquote>

while I understand the value of populating `self.resource`, if vCD replies with an error, the local cache might have inaccurate information anyways. It would be easier just to return the result of `put_resource` and not attempt to update `self.resource` by default
</blockquote></details>

Okay. Changing the code to simply return self.resource.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167)*

pacogomez:
:lgtm:

---

Reviewed 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/167#-:-L2TXBB6uAzB5mzkKMes:bnfp4nl)*

PR #196, opened 2018-01-29
VCDA-397: Refactor constructors to raise Error when both href and resource are None.

sahithi:
Currently, all pyvcloud Entity constructors expect one of href and resource variables to be populated. When both href and resource arguments are None, it fails at vCD rest-api calls instead of failing at pyvcloud level.

This change-set ensure that all constructors raise exception when both href and resource are found to be None.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/196)

### DIFF
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -50,6 +50,8 @@ class Org(object):
 
         """
         self.client = client
+        if href is None and resource is None:
+            raise TypeError("Org initialization failed as arguments are invalid or None")
         self.href = href
         self.resource = resource
         if resource is not None:
--- a/pyvcloud/vcd/org.py
+++ b/pyvcloud/vcd/org.py
@@ -51,7 +51,8 @@ class Org(object):
         """
         self.client = client
         if href is None and resource is None:
-            raise TypeError("Org initialization failed as arguments are invalid or None")
+            raise TypeError("Org initialization failed as arguments"
+                            " are either invalid or None")
         self.href = href
         self.resource = resource
         if resource is not None:
--- a/pyvcloud/vcd/pvdc.py
+++ b/pyvcloud/vcd/pvdc.py
@@ -29,6 +29,8 @@ class PVDC(object):
             representation of the entity.
         """
         self.client = client
+        if href is None and resource is None:
+            raise TypeError("PVDC initialization failed as arguments are invalid or None")
         self.href = href
         self.resource = resource
         if resource is not None:
--- a/pyvcloud/vcd/pvdc.py
+++ b/pyvcloud/vcd/pvdc.py
@@ -30,7 +30,8 @@ class PVDC(object):
         """
         self.client = client
         if href is None and resource is None:
-            raise TypeError("PVDC initialization failed as arguments are invalid or None")
+            raise TypeError("PVDC initialization failed as arguments"
+                            " are either invalid or None")
         self.href = href
         self.resource = resource
         if resource is not None:
--- a/pyvcloud/vcd/role.py
+++ b/pyvcloud/vcd/role.py
@@ -32,7 +32,8 @@ class Role(object):
         """
         self.client = client
         if href is None and resource is None:
-            raise TypeError("Role initialization failed as arguments are invalid or None")
+            raise TypeError("Role initialization failed as arguments"
+                            " are either invalid or None")
         self.href = href
         self.resource = resource
         if resource is not None:
--- a/pyvcloud/vcd/role.py
+++ b/pyvcloud/vcd/role.py
@@ -31,6 +31,8 @@ class Role(object):
 
         """
         self.client = client
+        if href is None and resource is None:
+            raise TypeError("Role initialization failed as arguments are invalid or None")
         self.href = href
         self.resource = resource
         if resource is not None:
--- a/pyvcloud/vcd/system.py
+++ b/pyvcloud/vcd/system.py
@@ -33,7 +33,8 @@ class System(object):
         """
         self.client = client
         if admin_href is None and admin_resource is None:
-            raise TypeError("System initialization failed as arguments are invalid or None")
+            raise TypeError("System initialization failed as arguments"
+                            " are either invalid or None")
         self.admin_href = admin_href
         self.admin_resource = admin_resource
         if admin_resource is not None:
--- a/pyvcloud/vcd/system.py
+++ b/pyvcloud/vcd/system.py
@@ -32,6 +32,8 @@ class System(object):
 
         """
         self.client = client
+        if admin_href is None and admin_resource is None:
+            raise TypeError("System initialization failed as arguments are invalid or None")
         self.admin_href = admin_href
         self.admin_resource = admin_resource
         if admin_resource is not None:
--- a/pyvcloud/vcd/vapp.py
+++ b/pyvcloud/vcd/vapp.py
@@ -33,6 +33,8 @@ class VApp(object):
     def __init__(self, client, name=None, href=None, resource=None):
         self.client = client
         self.name = name
+        if href is None and resource is None:
+            raise TypeError("VApp initialization failed as arguments are invalid or None")
         self.href = href
         self.resource = resource
         if resource is not None:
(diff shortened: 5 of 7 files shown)


hodgesrm:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[pyvcloud/vcd/org.py, line 53 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-L430ohLwf0NduXXtNw0:-L430ohLwf0NduXXtNw1:b-ohix2n) ([raw file](https://github.com/vmware/pyvcloud/blob/a9f69b10ed0bbeb907e32632d1704c4f7be1b3af/pyvcloud/vcd/org.py#L53)):*
>      if href is None and resource is None:
>             raise TypeError("Org initialization failed as arguments"
>                             " are either invalid or None")
>   

I do not believe this fully solves problems with incomplete initialization.  I logged a bug (https://jira.eng.vmware.com/browse/VCDA-479) that illustrated how fetching org resources using client.get_org_by_name resulted in an incompletely loaded Org class that could not find VDCs.  In this case to get the reload to work I had to alter org.py  as follows: 

     def get_vdc(self, name):
-        if self.resource is None:
+        if self.resource is None or self.resource == '':
             self.resource = self.client.get_resource(self.href)

Assuming I'm reading the code correctly your initialization may be subject to the same problem.  It's not just initial creation but also ensuring that we reload properly when required. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196)*

hodgesrm:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[pyvcloud/vcd/org.py, line 53 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-L430ohLwf0NduXXtNw0:-L433bqxDvS3WkfWqzSo:b-5wy3bn) ([raw file](https://github.com/vmware/pyvcloud/blob/a9f69b10ed0bbeb907e32632d1704c4f7be1b3af/pyvcloud/vcd/org.py#L53)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

>      if href is None and resource is None:
>             raise TypeError("Org initialization failed as arguments"
>                             " are either invalid or None")
>   

I do not believe this fully solves problems with incomplete initialization.  I logged a bug (https://jira.eng.vmware.com/browse/VCDA-479) that illustrated how fetching org resources using client.get_org_by_name resulted in an incompletely loaded Org class that could not find VDCs.  In this case to get the reload to work I had to alter org.py  as follows: 

     def get_vdc(self, name):
-        if self.resource is None:
+        if self.resource is None or self.resource == '':
             self.resource = self.client.get_resource(self.href)

Assuming I'm reading the code correctly your initialization may be subject to the same problem.  It's not just initial creation but also ensuring that we reload properly when required. 
</blockquote></details>

Ouch! The diff did not show correctly. It was supposed to show that I had to add "or self.resource='':"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196)*

rocknes:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[pyvcloud/vcd/org.py, line 53 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-L430ohLwf0NduXXtNw0:-L45am-PE6tiCTfip5xD:bqn4hfl) ([raw file](https://github.com/vmware/pyvcloud/blob/a9f69b10ed0bbeb907e32632d1704c4f7be1b3af/pyvcloud/vcd/org.py#L53)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

Ouch! The diff did not show correctly. It was supposed to show that I had to add "or self.resource='':"
</blockquote></details>

I think the issue you are talking about is caused by the fact that we are replacing Resource with Records. Well Known End points return Records which are trimmed down version of the actual resources.

We need to document the return types for functions in client.py and other files (which are missing documentation). I would attribute the bug you filed as improper use of api.

---

*[pyvcloud/vcd/vapp.py, line 33 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-L45a69O_9IZt05nWBuC:-L45a69O_9IZt05nWBuD:b-ekcftl) ([raw file](https://github.com/vmware/pyvcloud/blob/a9f69b10ed0bbeb907e32632d1704c4f7be1b3af/pyvcloud/vcd/vapp.py#L33)):*
> ```Python
> 
> class VApp(object):
>     def __init__(self, client, name=None, href=None, resource=None):
> ```

Do we really need the field 'name'?
Whenever the resource is loaded, name would be part of the xml.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196)*

rocknes:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[pyvcloud/vcd/org.py, line 53 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-L430ohLwf0NduXXtNw0:-L45bgHiRGWgvmZ-wzMf:bu2fq0r) ([raw file](https://github.com/vmware/pyvcloud/blob/a9f69b10ed0bbeb907e32632d1704c4f7be1b3af/pyvcloud/vcd/org.py#L53)):*

I think the issue you are talking about is caused by the fact that we are replacing Resource with Records. Well Known End points return Records which are trimmed down version of the actual resources.

We need to document the return types for functions in client.py and other files (which are missing documentation). I would attribute the bug you filed as improper use of api.
</blockquote></details>

I believe we should raise an Exception (new type?) which signals the client that this is a non recoverable error and they shouldn't catch it. If the object is initialized with incomplete data, there is no way we can salvage the situation!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196)*

hodgesrm:
:lgtm:

---

Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions.

---

*[pyvcloud/vcd/org.py, line 53 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-L430ohLwf0NduXXtNw0:-L47g6JE6rFveHSN0ALF:bcvbf69) ([raw file](https://github.com/vmware/pyvcloud/blob/a9f69b10ed0bbeb907e32632d1704c4f7be1b3af/pyvcloud/vcd/org.py#L53)):*

I believe we should raise an Exception (new type?) which signals the client that this is a non recoverable error and they shouldn't catch it. If the object is initialized with incomplete data, there is no way we can salvage the situation!
</blockquote></details>

Sahithi and I reviewed the code and underlying data model face-to-face and I agree that my case was improper usage .  The real "bug" in that case is that (a) the method did not indicate the return value very clearly and (b) that we don't have external documentation the object model.   I'll update my JIRA(s) to reflect that.  Meanwhile this is good to go. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-:-L47x5mz8uy4TTAV-Zoa:bnfp4nl)*

rocknes:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196)*

sahithi:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[pyvcloud/vcd/vapp.py, line 33 at r1](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-L45a69O_9IZt05nWBuC:-L4Hf2oq1n6wN7ywiXgm:bking3o) ([raw file](https://github.com/vmware/pyvcloud/blob/a9f69b10ed0bbeb907e32632d1704c4f7be1b3af/pyvcloud/vcd/vapp.py#L33)):*

Do we really need the field 'name'?
Whenever the resource is loaded, name would be part of the xml.
</blockquote></details>

Out of current scope of the change-set. I will let Sompa know of this and she can evaluate if there is a need and can do necessary changes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196)*

sompa:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/vmware/pyvcloud/196#-:-L4Hvop6-21RXaYOt2Kr:bnfp4nl)*

PR #209, opened 2018-02-20
added function to get the nat rules of an edge gateway by name

floriankammermann:
added a function to get the nat rules of an edge gateway by name.  
The function calls list_edge_gateways.
It checks the received array, if the requested (name) edge gateway is included.
If yes, it tries to get the nat rules.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/209)

### DIFF
--- a/tests/vcd_edge_gateway.py
+++ b/tests/vcd_edge_gateway.py
@@ -13,8 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import unittest
-
 from pyvcloud.vcd.org import Org
 from pyvcloud.vcd.test import TestCase
 from pyvcloud.vcd.vdc import VDC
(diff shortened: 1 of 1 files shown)


sompa:
Hi Florian,

You have to fix the code to avoid flake8 errors https://travis-ci.org/vmware/pyvcloud/builds/344096397?utm_source=github_status&utm_medium=notification 

Also please go through the contributing guide here https://github.com/sompa/pyvcloud/blob/master/CONTRIBUTING.md#code-style 
for the code style and standards that is being followed for this project. 

We can review your pull request once the above two requirements are met. 

Thanks
Sompa

pyvcloud/vcd/vdc.py
```diff
@@ -373,6 +373,42 @@ def list_edge_gateways(self):
...
+    def get_nat_rules(self, edge_gateway_name):
+        """Request the nat rules of an edge gateway.
+
+        :param edge_gateway_name: (str): The name of the edge gateway.
+
+        :return: A list of :class:`lxml.objectify.StringElement' objects
+            representing the nat rules.
+
+        :raises: Exception: if there is no edge gateway with the provided name.
+        :raises: Exception: if there are no nat rules on the edge gateway
```
rocknes: I guess it's possible to have an edge gateway without any NAT rules configured on it.  Keeping that in mind, I don't think it's correct to throw an exception if no NAT rules are found on the edge gateway. Rather we should just return an empty list back.

pyvcloud/vcd/vdc.py
```diff
@@ -373,6 +373,42 @@ def list_edge_gateways(self):
...
+            with the provided name.
+        """
+        edge_gateways = self.list_edge_gateways()
+        nat_rules = []
+        for edge_gateway in edge_gateways:
```
rocknes: This piece of code that retrieves the full representation of an edge gateway identified by name, should be refactored out in a separate function. I am sure that this functionality will be a recurring theme while accessing/updating other services/properties of an edge gateway.

pyvcloud/vcd/vdc.py
```diff
@@ -373,6 +373,42 @@ def list_edge_gateways(self):
...
+        :param edge_gateway_name: (str): The name of the edge gateway.
+
+        :return: A list of :class:`lxml.objectify.StringElement' objects
+            representing the nat rules.
+
+        :raises: Exception: if there is no edge gateway with the provided name.
+        :raises: Exception: if there are no nat rules on the edge gateway
+            with the provided name.
+        """
+        edge_gateways = self.list_edge_gateways()
```
rocknes: I would name the variable as edge_gateway_records to avoid any confusion.

pyvcloud/vcd/vdc.py
```diff
@@ -373,6 +373,42 @@ def list_edge_gateways(self):
...
+                           'EdgeGatewayServiceConfiguration') and \
+                   hasattr(edge_gateway_full.Configuration.
+                           EdgeGatewayServiceConfiguration,
+                           'NatService'):
+                    for nat_rule in edge_gateway_full.Configuration.\
+                            EdgeGatewayServiceConfiguration.NatService.NatRule:
+                        nat_rules.append(nat_rule)
+                if len(nat_rules) > 0:
+                    return nat_rules
+                else:
```
rocknes: Please see my comment, regarding throwing exceptions, at the top of this function.

tests/vcd_edge_gateway.py
```diff
@@ -29,3 +29,20 @@ def test_001_list_edge_gateways(self):
...
+        vdc = VDC(self.client, resource=v)
+        nat_rules = vdc.get_nat_rules(self.config['vcd']['edge_gateway'])
+        assert len(nat_rules) > 0
+
+    def test_003_get_nat_rules_edge_gateway_not_existent(self):
+        logged_in_org = self.client.get_org()
+        org = Org(self.client, resource=logged_in_org)
+        v = org.get_vdc(self.config['vcd']['vdc'])
+        vdc = VDC(self.client, resource=v)
+        with self.assertRaisesRegex(Exception, 'Edge Gateway not found \''+self.config['vcd']['edge_gateway']+'_not_existent\'') as exception:
```
rocknes: Ideally we should have a specific exception for edge gateway not found error. Which we would catch in this method to mark this test as passed.
Since, such an exception is not currently available to us, we can run the above code block in a try-catch block, and verify that the caught exception's message matches the expected message.

We don't need to print the actual exception content on terminal.

floriankammermann: as I understand https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex verifies the text of the exception message.

floriankammermann:
I did something wrong with the upstream sync. Will open a new pull request with the new adjusted code.

PR #239, opened 2018-05-04
[VCDA-574] Implement support for ovf:chunkSize while uploading vAppTemplates to a catalog.

rocknes:
https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_1.1.0.pdf
Section 7.1 talks about File References.

'''"Files referenced from the reference part may be split into chunks to accommodate file size restrictions on certain file systems. Chunking shall be indicated by the presence of the ovf:chunkSize attribute; the
value of ovf:chunkSize shall be the size of each chunk, except the last chunk, which may be smaller.'''

Without adding support for this parameter, trying to upload ova files with this attribute would cause upload failures.

In this commit,
* Added support for ovf:chunkSize while uploading vAppTemplates to catalog.
* Refactored the upload/download functions without changing their external apis.
* Optimized enable-download stage during catalog item downloads.

Testing Done:
1. Added a new ova file that has ovf:chunkSize param in it.
2. Wrote system tests to test upload/download operations on catalog using a vanilla ova file and the special ova file mentioned above.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)


hodgesrm:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[a discussion](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg3RGIA9V-BUivW7lg:-LBg3RGIA9V-BUivW7lh:b-2tudpo) (no related file):*
This change is pretty big and involves detailed behavior that that is difficult to verify visually.  The commit needs an accompanying system test that shows we (a) didn't break existing behavior and (b) that the new chunked behavior is correct.  That will complete what looks like an overall very nice cleanup. 

One suggestion on the test--we'll need a large OVA to check.  I suggest using the Photon OS VM and downloading it to a known location.  You can get it with 'wget http://dl.bintray.com/vmware/photon/2.0/GA/ova/photon-custom-hw11-2.0-304b817.ova'.  If the OVA is already present and the SHA1 matches,  just use that. The test will can then run any number of times times without waiting for a new download.  In fact this might be a nice thing to put in Environment so we don't have to repeat that logic across multiple tests.  

I can also change the samples to behave similarly.  The YAML file parameters could include the URL as well as the MD5 sum.  If it's http[s] we download.  If it's just a file path we use it, so there could be good flexibility.

From reading the code it seems you can set the chunksize to trigger chunking on upload and download even for smaller VMs.  That should exercise the transmission logic properly. 

---

*[a discussion](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg70Dg9lxmjCBYAlQG:-LBg70Dg9lxmjCBYAlQH:b-xcarka) (no related file):*
Docstring content is outstanding.  Great model for the rest of the API surface. 

---

*[pyvcloud/vcd/org.py, line 228 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg6rWw3Sy8IqpH0WHI:-LBg6rWw3Sy8IqpH0WHJ:b-7ir5i2) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L228)):*
> ```Python
> 
>     def _is_enable_download_required(self, entity_resource, item_type):
>         """Helper method to determines need for download enablement.
> ```

"determines" should be "determine"

---

*[pyvcloud/vcd/org.py, line 280 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg9YH5-37swGipTebB:-LBg9YH5-37swGipTebC:b-wz0a8e) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L280)):*
> ```Python
>                               callback=None,
>                               task_callback=None):
>         """Downloads an item from a catlog into a local file.
> ```

"catlog" should be "catalog"

---

*[pyvcloud/vcd/org.py, line 419 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg9z8cEoSskFLs8Iym:-LBg9z8cEoSskFLs8Iyn:bkd64xs) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L419)):*
> ```Python
>         :param file_name: (str): The name of the media file on local disk
>             which will be uploaded.
>         :param item_name: (str): This param let's us rename the media file
> ```

"let's" should be "lets"

---

*[pyvcloud/vcd/org.py, line 614 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBgAhS6EbxpHJAfc33k:-LBgAhS6EbxpHJAfc33l:b-4h19yj) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L614)):*
> ```Python
>             progress of the upload operation.
> 
>         :return: Number of bytes uploade to the uri.
> ```

"uploade" should be "uploaded"

---

*[pyvcloud/vcd/org.py, line 637 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBgAoAkBRikdxbBfj97:-LBgAoAkBRikdxbBfj98:b-4h19yj) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L637)):*
> ```Python
>             progress of the upload operation.
> 
>         :return: Number of bytes uploade to the uri.
> ```

"uploade" should be "uploaded"

---

*[pyvcloud/vcd/org.py, line 674 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBgAvMtA224RM_dKC5O:-LBgAvMtA224RM_dKC5P:b-4h19yj) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L674)):*
> ```Python
>             progress of the upload operation.
> 
>         :return: Number of bytes uploade to the uri.
> ```

"uploade" should be "uploaded"

---


*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)*

rocknes:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 7 unresolved discussions.

---

*[a discussion](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg3RGIA9V-BUivW7lg:-LBgUSBS9wQZC7YPTQzb:b-d25ev8) (no related file):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

This change is pretty big and involves detailed behavior that that is difficult to verify visually.  The commit needs an accompanying system test that shows we (a) didn't break existing behavior and (b) that the new chunked behavior is correct.  That will complete what looks like an overall very nice cleanup. 

One suggestion on the test--we'll need a large OVA to check.  I suggest using the Photon OS VM and downloading it to a known location.  You can get it with 'wget http://dl.bintray.com/vmware/photon/2.0/GA/ova/photon-custom-hw11-2.0-304b817.ova'.  If the OVA is already present and the SHA1 matches,  just use that. The test will can then run any number of times times without waiting for a new download.  In fact this might be a nice thing to put in Environment so we don't have to repeat that logic across multiple tests.  

I can also change the samples to behave similarly.  The YAML file parameters could include the URL as well as the MD5 sum.  If it's http[s] we download.  If it's just a file path we use it, so there could be good flexibility.

From reading the code it seems you can set the chunksize to trigger chunking on upload and download even for smaller VMs.  That should exercise the transmission logic properly. 
</blockquote></details>

I agree, I will add a test for this.

---

*[pyvcloud/vcd/org.py, line 228 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg6rWw3Sy8IqpH0WHI:-LBgUXXn30T97DW4e5cn:b-886gsx) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L228)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

"determines" should be "determine"
</blockquote></details>

Fixed

---

*[pyvcloud/vcd/org.py, line 280 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg9YH5-37swGipTebB:-LBgUYOu3YqqjsXCU7GT:b-886gsx) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L280)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

"catlog" should be "catalog"
</blockquote></details>

Fixed

---

*[pyvcloud/vcd/org.py, line 419 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg9z8cEoSskFLs8Iym:-LBgUZU94nD0Ucets4oU:b-886gsx) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L419)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

"let's" should be "lets"
</blockquote></details>

Fixed

---

*[pyvcloud/vcd/org.py, line 614 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBgAhS6EbxpHJAfc33k:-LBgU_Ek3wXUvkYK2pwV:b-886gsx) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L614)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

"uploade" should be "uploaded"
</blockquote></details>

Fixed

---

*[pyvcloud/vcd/org.py, line 637 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBgAoAkBRikdxbBfj97:-LBgU_xx8OdiJSVnoxBn:b-886gsx) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L637)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

"uploade" should be "uploaded"
</blockquote></details>

Fixed

---

*[pyvcloud/vcd/org.py, line 674 at r1](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBgAvMtA224RM_dKC5O:-LBgUack9sztzAgcL2iz:b-886gsx) ([raw file](https://github.com/vmware/pyvcloud/blob/ab98baf763d9d6d0989ec96f7b0049b1afca2a5b/pyvcloud/vcd/org.py#L674)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

"uploade" should be "uploaded"
</blockquote></details>

Fixed

---


*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)*

rocknes:
Review status: 0 of 4 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/vmware/pyvcloud/239#-LBg3RGIA9V-BUivW7lg:-LCh-ZfuAA9t7uoIeBOX:bufim0i) (no related file):*

I agree, I will add a test for this.
</blockquote></details>

Added tests.

---


*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)*

hodgesrm:
Reviewed 2 of 3 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[system_tests/catalog_tests.py, line 33 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCje-Rs3skXcqVa6V-X:-LCje-Rs3skXcqVa6V-Y:bncpnuo) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L33)):*
> ```Python
>         """Setup the catalog and templates for the tests in this module.
> 
>         Create a catalog and upload two templates to the catalog.
> ```

this is the only case that actually uploads files, it appears. Yet it does not actually do the upload if the items already exist.  Seems as if this means much of the time we might not actually test upload.  If I'm reading the code correctly there should be a case that checks upload each time the test runs. 

---

*[system_tests/catalog_tests.py, line 84 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCjcmrz3ISsUGomMqiA:-LCjcmrz3ISsUGomMqiB:b-gfyfe6) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L84)):*
> ```Python
>         into the spool area(transfer folder) but it doesn't wait for the
>         import of the template as catalog item to finish.
> ```

Awesome comments.  If you run the test two times in rapid succession will you somehow collide with the previous uploaded catalog item?  

---

*[system_tests/catalog_tests.py, line 189 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCjdRp678tKvNmYxbt_:-LCjdRp678tKvNmYxbta:b-xyvz5g) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L189)):*
> ```Python
> 
> 
> if __name__ == '__main__':
> ```

I was expecting to see more cases.  Are there other upload/download cases?

---


*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)*

rocknes:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[system_tests/catalog_tests.py, line 33 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCje-Rs3skXcqVa6V-X:-LCk477L12Xi6W8OyD85:b-m9revn) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L33)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

this is the only case that actually uploads files, it appears. Yet it does not actually do the upload if the items already exist.  Seems as if this means much of the time we might not actually test upload.  If I'm reading the code correctly there should be a case that checks upload each time the test runs. 
</blockquote></details>

I was following the pattern we have established that don't create/upload things and reuse them if possible.
I can surely relax that for this case and create a new catalog every time this test is run, which will guarantee that the upload is triggered. :)

---

*[system_tests/catalog_tests.py, line 84 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCjcmrz3ISsUGomMqiA:-LCjrDMwBqvDfZY3N1yv:b-j2r8hl) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L84)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

Awesome comments.  If you run the test two times in rapid succession will you somehow collide with the previous uploaded catalog item?  
</blockquote></details>

The second call will fail.

Uploading an item to a catalog involves the following steps,
1. Do a POST call on vCD to let vCD know that you are trying to upload a catalog item (media / template)
2. Wait for vCD to create proper transfer folder corresponding to the catalog item. (Unfortunately this is not implemented via tasks in vCD, so we need to monitor the catalog item resource and wait for the url of catalog item to appear in the response).
3. Upload the bits to the transfer folder
4. Wait for vCD to finish importing whatever you have uploaded to transfer folder into the ESX host.

Step 1 is what guarantees that the second upload call will fail with the error message "Item with same name exists in the catalog"

---

*[system_tests/catalog_tests.py, line 189 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCjdRp678tKvNmYxbt_:-LCk3d76C27ZMqAwxDX1:b-m6bsb3) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L189)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

I was expecting to see more cases.  Are there other upload/download cases?
</blockquote></details>

I was more focused on testing the ovf upload download path. The other upload/download path is the media. But that path is almost identical to download/upload of templates without ovf:chunkSize.  I will write more tests related to catalogs as part of VCDA-586

---


*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)*

hodgesrm:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[system_tests/catalog_tests.py, line 33 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCje-Rs3skXcqVa6V-X:-LCkjcmfCkXA6vR_GU7W:bfoeydg) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L33)):*

Also forgot to mention that, if you don't turn on developer mode the catalog where the templates are uploaded will be deleted as part of cleanup. So next time when these test are run, they will *create* the catalog again and then upload the templates. So under normal runs, the upload code path will be tested all the time.
</blockquote></details>

Since this is a test of upload we should ensure we do upload.  It should be baked into the test case; otherwise you may pass software that actually has failures. 

---

*[system_tests/catalog_tests.py, line 84 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCjcmrz3ISsUGomMqiA:-LCkjsS13AbnKOyW0v0L:b-dpwyv3) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L84)):*

The second call will fail.

Uploading an item to a catalog involves the following steps,
1. Do a POST call on vCD to let vCD know that you are trying to upload a catalog item (media / template)
2. Wait for vCD to create proper transfer folder corresponding to the catalog item. (Unfortunately this is not implemented via tasks in vCD, so we need to monitor the catalog item resource and wait for the url of catalog item to appear in the response).
3. Upload the bits to the transfer folder
4. Wait for vCD to finish importing whatever you have uploaded to transfer folder into the ESX host.

Step 1 is what guarantees that the second upload call will fail with the error message "Item with same name exists in the catalog"
</blockquote></details>

I think you should add logic to ensure it completes the upload if this would avoid a build failing due to running the case quickly. Your choice whether that's necessary.  

---


*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)*

rocknes:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[system_tests/catalog_tests.py, line 33 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCje-Rs3skXcqVa6V-X:-LClE15u62ZYkHHvNQin:b-x3bcfn) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L33)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

Since this is a test of upload we should ensure we do upload.  It should be baked into the test case; otherwise you may pass software that actually has failures. 
</blockquote></details>

Sure, I will change the catalog name to random string.

---

*[system_tests/catalog_tests.py, line 84 at r3](https://reviewable.io/reviews/vmware/pyvcloud/239#-LCjcmrz3ISsUGomMqiA:-LClEFJi80OZ51a4HPMt:b-vefab2) ([raw file](https://github.com/vmware/pyvcloud/blob/78bcee2f8521ca4d11e2e3edada9f2cc20eb333c/system_tests/catalog_tests.py#L84)):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

I think you should add logic to ensure it completes the upload if this would avoid a build failing due to running the case quickly. Your choice whether that's necessary.  
</blockquote></details>

If I change the setup test to use a new catalog everytime the test is run, there will be no race conditions. I guess that solves both of our problems :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239)*

hodgesrm:
:lgtm:

---

Reviewed 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239#-:-LCoa7qx36vEbDY6NVfH:bnfp4nl)*

sahithi:
:lgtm:

---

Reviewed 2 of 3 files at r2, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/239#-:-LD7rJeJ1ORvcZeUbx1n:bnfp4nl)*

PR #287, opened 2018-09-17
[VDP-2083, VCDA-721] Implement migrate-vms from one resource pool to another

ckolovson:
This change is to implement migrate-vms from one resource pool to another. Testing includes migrating vms from one RP to another using a specified target RP, and migrating back using a non-specified target RP. 

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/287)


ckolovson:
Implemented migrate-vms from one resource pool to another, including system test. Also included a minor bug fix regarding attach VC.

PR #294, opened 2018-09-21
platform: Do not prefix or suffix URL parameters

ckolovson:
- Do not prepend 'https://' to a VC / NSX host argument for the attach VC / NSX
   workflow if the supplied parameter is already a URL

 - Do not append ':443' to a VC / NSX host argument for the attach VC / NSX
   workflow if the supplied parameter is already a URL

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/294)


ckolovson:
Folding this into PR #287

PR #304, opened 2018-10-19
[VCDA-786] Additional fix for failures during template uploads to catalog. 

rocknes:
This PR is addendum to https://github.com/vmware/pyvcloud/pull/303

If vCD servers are behind a load balancer, the TCP connections are generally closed as soon as the first response is sent out by the load balancer. If we pump too much data in a short burst to such a system, we run the risk of getting back 416 instead of 200. Additionally we run the risk of getting the error
``urllib3.exceptions.ProtocolError: ('Connection aborted.', ConnectionResetError(10054, 'An existing connection was forcibly closed by the remote host', None, 10054, None))``

In this PR, we address these two concerns, by 
1. Introducing a retry mechanism in case we get back non 200 status on a PUT call for uploading content.
2. Increasing the upload chunk size from 1 MB to 10 MB.
3. Introducing delay between successive PUT calls if we detect that the server is closing the TCP connections.

Tested this fix by uploading and downloading templates to/from a vCD cluster sitting behind a load balancer. Without these fixes the upload never succeeded. With the fix both upload and download are working fine. Also tested the new code against a stand alone vCD server and to make sure that no regressions were introduced.

Signed by: Aritra Sen<aritra.iitkgp@gmail.com

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/304)

### DIFF
--- a/pyvcloud/vcd/client.py
+++ b/pyvcloud/vcd/client.py
@@ -35,7 +35,8 @@ from pyvcloud.vcd.exceptions import AccessForbiddenException, \
     MultipleRecordsException, NotAcceptableException, NotFoundException, \
     OperationNotSupportedException, RequestTimeoutException,\
     TaskTimeoutException, UnauthorizedException, UnknownApiException, \
-    UnsupportedMediaTypeException, VcdException, VcdTaskException  # NOQA
+    UnsupportedMediaTypeException, VcdException, VcdResponseException, \
+    VcdTaskException  # NOQA
 
 
 SIZE_1MB = 1024 * 1024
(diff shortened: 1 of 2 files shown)


pyvcloud/vcd/client.py
```diff
@@ -868,6 +899,40 @@ def _response_code_to_exception(sc, request_id, objectify_response):
 
         raise UnknownApiException(sc, request_id, objectify_response)
 
+    def _log_request_response(self,
```
sahithi: Docstring for this.

pyvcloud/vcd/client.py
```diff
@@ -868,6 +899,40 @@ def _response_code_to_exception(sc, request_id, objectify_response):
...
+        if self._log_headers:
+            self._logger.debug('Request headers: %s' %
+                               response.request.headers)
+
+        if self._log_bodies and request_body is not None:
+            if isinstance(request_body, str):
+                body = request_body
+            else:
+                body = request_body.decode(self.fsencoding)
+            self._logger.debug('Request body: %s' % body)
```
sahithi: What will be the body for template upload?

PR #333, opened 2018-12-03
VP-1243: Modify an IP range of a subnet in external network

kousgy123:
VP-1243: Modify an IP range of a subnet in external network

This changelist involves the modification of a IP range in the subnet with the new IP range.
Testing done:
1. flake8 command succeeded
2. Test case passed, Ip Range modified in the subnet.

- (Optional) Names of reviewers using @ sign + name

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/333)


kousgy123:
1. EntityNotFoundException is thrown when IP range is not present.
2. Helper method added for the common code.

kousgy123:
removed both redundant return statement.
modified the method to private.

PR #346, opened 2018-12-20
VP1259: [PySdk] Detach port groups from an external network

pandeys:
[PySdk] Detach port groups from an external network

This code has both functionality and test case for detaching port group
from external network.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/346)

### DIFF
--- a/pyvcloud/vcd/external_network.py
+++ b/pyvcloud/vcd/external_network.py
@@ -225,9 +225,9 @@ class ExternalNetwork(object):
             if ip_scope.Gateway == gateway_ip:
                 for exist_ip_range in ip_scope.IpRanges.IpRange:
                     if exist_ip_range.StartAddress == \
-                       old_ip_addrs[0] and \
-                       exist_ip_range.EndAddress \
-                       == old_ip_addrs[1]:
+                            old_ip_addrs[0] and \
+                            exist_ip_range.EndAddress \
+                            == old_ip_addrs[1]:
                         exist_ip_range['StartAddress'] = \
                             E.StartAddress(new_ip_addrs[0])
                         exist_ip_range['EndAddress'] = \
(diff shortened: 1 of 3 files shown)


pyvcloud/vcd/external_network.py
```diff
@@ -315,3 +315,42 @@ def __create_vimobj_ref(self, vc_href, pg_moref, pg_type):
...
+                "Either vCenter Server name is none or portgroup name is none")
+
+        vc_record = platform.get_vcenter(vim_server_name)
+        vc_href = vc_record.get('href')
+        pg_moref_types = \
+            platform.get_port_group_moref_types(vim_server_name,
+                                                port_group_name)
+
+        vim_port_group_refs = \
+            ext_net['{' + NSMAP['vmext'] + '}VimPortGroupRefs']
```
kousgy123: Need to handle the condition when the single port group is present that is "VimPortGroupRef".

Check in the VCD UI when single port group is present and you are trying to delete the same.

system_tests/helpers/portgroup_helper.py
```diff
@@ -49,8 +49,40 @@ def get_available_portgroup_name(self, vim_server_name, portgroupType):
...
+        """
+        name_filter = ('vcName', vim_server_name)
+
+        query = self.client.get_typed_query(
+            ResourceType.PORT_GROUP.value,
+            query_result_format=QueryResultFormat.RECORDS,
+            equality_filter=name_filter)
+        pgroup_name = ''
+        for record in list(query.execute()):
+            if record.get('networkName') == ext_net_name:
```
kousgy123: Please check if below works:
if record.get('networkName') == ext_net_name and
       not record.get('name').startswith('vxw-'):

i fell portgroup check is not required.

PR #367, opened 2019-01-08
Add several methods to vdc, vapp and vm

dwettstein:
This PR adds 7 methods:
- VDC:
    - get_resource_admin()
    - query_vm_by_name(name)
    - get_vc()
- ~~VApp~~:
    - ~~set_name(name)~~
- VM:
    - discard_suspended_state()
    - ~~set~~edit_name(name)
    - ~~set~~edit_hostname(hostname)

Unfortunately, I was not able to write some tests for those, as I don't have a local vCD instance and our environments don't have internet access for installing the dependencies. However, as the code is not very special that shouldn't be a problem, one can add the tests later. Additionally, we are already using it in production.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/367)


pandeys:
Please sync your code because a lot of portion is already implemented. Remaining code, you should submit with test case in system_tests folder.

shashim22:
Showing conflict in pyvcloud/vcd/client.py. Please resolve it.

PR #472, opened 2019-03-01
Update test config.yml with missing parameters

cdivitotawela:
* Unit tests vcd_org and vcd_system fails as they look for parameters config['vcd']['org_name] and config['vcd']['org_full_name'] which are missing. Added these parameters with sensible values
* Delete org test fails. Updated to get system resource with admin resource

Signed-off-by: Chaminda Divitotawela <cdivitotawela@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/472)


pandeys:
We will remove tests folder and all test cases will be executed only from system_tests folder. Please don't contribute test cases in this folder.

PR #480, opened 2019-03-05
Add mac_address to VM NIC listing

mr00wka:
Add `mac_address` to dict produced by vm.list_nics() method

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/480)


mr00wka:
Is there a point of testing it if it is only set when `NetworkConnectionSection.NetworkConnection` actually has this attribute? Similarly as with `IpAddress`

PR #634, opened 2019-12-04
Preserve session's data when connecting for further usages

lrivallain:
Preserve session's data when connecting for further usages

Set a `client.session` member based on the `objectify.fromstring(response.content)` from POST request to the /sessions. This enable to retrieve the session content (like current user's role) from the client object when needed.

Ex: Get user's role:
```python
# Get role for user
user_role = user_client.session.get('roles')
```

Backward compatibility should be preserved as the session is also returned directly when calling the `client.rehydrate_from_token(token)` method.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/634)

### DIFF
--- a/pyvcloud/vcd/client.py
+++ b/pyvcloud/vcd/client.py
@@ -966,13 +966,13 @@ class Client(object):
                 else:
                     raise VcdException('Login failed.')
 
-            session = objectify.fromstring(response.content)
-            self._session_endpoints = _get_session_endpoints(session)
+            self.session = objectify.fromstring(response.content)
+            self._session_endpoints = _get_session_endpoints(self.session)
 
             self._session = new_session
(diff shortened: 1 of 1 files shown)


lrivallain:
Hi @shashim22, thanks for the review and suggestions.

Sadly I do not have an appropriate environment to run `client_tests`.

lrivallain:
Hello shashim22,

I have made manual test already but I cannot confirm that it will not break something that I did not cover ;-)

However I use this change in my projects without having found a side effect.

Regards,

shashim22:
This has been taken care by the commit:
https://github.com/vmware/pyvcloud/commit/c909cb9e1df6ad39031da491096a458426dd7bcb

PR #660, opened 2020-05-12
Fix - disk list size

azzamsa:
- Fix listing disk: `disk` object has no `size` attribute, but `sizeMb`.

Related: https://github.com/vmware/vcd-cli/pull/527

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/660)

### DIFF
--- a/pyvcloud/vcd/utils.py
+++ b/pyvcloud/vcd/utils.py
@@ -549,8 +549,12 @@ def disk_to_dict(disk):
     result['name'] = disk.get('name')
     result['id'] = extract_id(disk.get('id'))
     result['status'] = disk.get('status')
-    result['size'] = humanfriendly.format_size(int(disk.get('size')))
-    result['size_bytes'] = disk.get('size')
+    if 'size' in disk.keys():
+        size_in_bytes = int(disk.get('size'))
+    else:
+        size_in_bytes = int(disk.get('sizeMb')) * (1024 * 1024)
+    result['size'] = humanfriendly.format_size(size_in_bytes)
+    result['size_bytes'] = size_in_bytes
     result['busType'] = disk.get('busType')
     result['busSubType'] = disk.get('busSubType')
     result['iops'] = disk.get('iops')


azzamsa:
Thanks for the input.

To get the api version, I need to do something like:

```python
    size_key = 'size'
    if client.get_api_version() < ApiVersion.VERSION_33.value:
        size_key = 'sizeMb'
```

because `disk` object doesn't contain any information about api version. So now `disk_to_dict` will need two parameters `client` and `disk`. 

Any cleaner solution?

azzamsa:
> We can first check 'sizeMb' if it doesn't have a valid value, read 'size'. Since this is a mandatory parameter, either of the two will have a valid value. This way, we can skip checking API version.

Done.

azzamsa:
the correct formula would be:
```
  size_in_bytes = int(disk.get('sizeMb')) * (1000 * 1000)
```

Tested and compared the `size` using vcloud director this lib.

azzamsa:
### In vcloud director

image

### using `1000 * 1000` (API result)
```
"size": "953 MB",
"size_bytes": "953000000",
```

### using `1024 * 1024` (API result)

```
"size": "999.29 MB",
"size_bytes": "999292928",
```

How?

azzamsa:
:lgtm:

azzamsa:
___
*[pyvcloud/vcd/utils.py, line 552 at r2](https://reviewable.io/reviews/vmware/pyvcloud/660#-M7kSzEJ4Nd8Q5v0mffR:-M7lGkEP2KCFy9DmxiFa:b-896fix) ([raw file](https://github.com/vmware/pyvcloud/blob/043b52ef22c4a23eeebd024b15a8fcfe05dbc47c/pyvcloud/vcd/utils.py#L552)):*
<details><summary><i>Previously, shashim22 wrote…</i></summary><blockquote>

Thanks for pointing it out, that was a typo. We should use 1024*1024 to be precise.  
In the IF block, size_in_bytes should be converted to int:
size_in_bytes = int(disk.get('size'))


</blockquote></details>

Done.

PR #671, opened 2020-06-05
Get the Org from the Client if find_link fails in instantiate_vapp()

juanfont:
`instantiate_vapp` in vdc.py assumes that the current vdc has an up link to the Org. This is not true when using an unprivileged account like vApp Author, that has certain access to public catalogues.

This patch allows to fetch the current Org from the client, which will be the one we are working with for this VDC (the unprivileged user does not have visibility on other tenants like the SYSTEM account)

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/671)


juanfont:
@ shashim22 can you have a look to this one? :)

Issue #682, opened 2020-07-07
Error 400 occurs when instantiating vApp from template with internal vApp network


Ptomerty:
Following up from https://github.com/vmware/vcd-cli/issues/534 since I read the source code and all roads led to [`instantiate_vapp()`](https://github.com/vmware/pyvcloud/blob/master/pyvcloud/vcd/vdc.py#L183).

Is there any way to instantiate a vApp (in this case, with multiple VMs) that already has an internal vApp network? Currently, the network settings default to looking for this network in the list of VDC networks, but the template has a vApp network, so obviously it won't be found. Thanks!

EDIT: It seems the error is that a template vApp network is incorrectly treated as a VDC network. Submitting PR to fix this issue.

junankar:
@Ptomerty I was looking for exact same issue today! Thanks for submitting a fix.

junankar:
I am also looking into creating vApp with just template information and vApp name (similar to webclient).

e.g.

<ns6:InstantiateVAppTemplateParams name="vAppName">
    <ns6:InstantiationParams/>
    <ns6:Source href="https://IP/api/vAppTemplate/vappTemplate-ID" id="urn:vcloud:vapptemplate:ID" name="vappTemplate-Name" type="application/vnd.vmware.vcloud.vAppTemplate+xml"/>
</ns6:InstantiateVAppTemplateParams>

Is there any way to skip specifying sourceitems ( SourcedItem  element of SourcedCompositionItemParamType)?

Following code is setting sourced_item:
vapp_template_params.append(sourced_item)

Ptomerty:
You should create your own issue for that. As far as I can tell, though, that's not possible; the sourced_item object seems to be the actual VM before it's created.

PR #683, opened 2020-07-08
Correctly ignore vApp networks when instantiating from template

Ptomerty:
When instantiating a new vApp from a template that has an existing vApp network, the current code attempts to look for this vApp network in the list of VDC networks, which results in an 400 error and EntityNotFoundException. To correct this, we skip the exception if we encounter a template network, which lets vCloud Director correctly handle the vApp network.

Fixes https://github.com/vmware/pyvcloud/issues/682.

Tested on an internal vCloud Director instance @ VMware, and verified that the changes made here correctly propagate to [`vcd-cli`](https://github.com/vmware/vcd-cli/) and [`ansible-module-vcloud-director`](https://github.com/vmware/ansible-module-vcloud-director).

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/683)

### DIFF
--- a/pyvcloud/vcd/vdc.py
+++ b/pyvcloud/vcd/vdc.py
@@ -263,6 +263,7 @@ class VDC(object):
 
         # If network is not specified by user then default to
         # vApp network name specified in the template
+        network_is_template_vapp_network = False
         template_networks = template_resource.xpath(
             '//ovf:NetworkSection/ovf:Network',
             namespaces={'ovf': NSMAP['ovf']})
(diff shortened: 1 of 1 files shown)


pyvcloud/vcd/vdc.py
```diff
@@ -263,6 +263,7 @@ def instantiate_vapp(self,
 
         # If network is not specified by user then default to
         # vApp network name specified in the template
+        network_is_template_vapp_network = False
```
shashim22, line 266: Could you please rename is to is_template_vapp_network?

PR #711, opened 2020-09-30
Add support to escape special characters in query filter for api v35.0+

rocknes:
For api v35.0+, vCD requires special characters viz. ( ) ; , to be escaped if used as value in query filter.
In _AbstractQuery, we escape the special characters in encoded filter value.

Additional changes, 
1. Replaced usage of quote_plus to quote, since quote_plus will replace spaces with + and thus can cause the filter value to distort.
2. Encoded all usage of qfilter in get_typed_query, since this field should be encoded by caller.

Testing done:
Ran all vcd-cli commands that invoke the affected methods at api v35.0 and made sure that all these commands returned expected result.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/711)

### DIFF
--- a/pyvcloud/vcd/api_extension.py
+++ b/pyvcloud/vcd/api_extension.py
@@ -82,9 +82,9 @@ class APIExtension(object):
         :raise MultipleRecordsException: if more than one service with the
             given name and namespace are found.
         """
-        qfilter = 'name==%s' % urllib.parse.quote_plus(name)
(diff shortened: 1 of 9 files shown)


shashim22 (approved):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/711)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #713, opened 2020-10-23
Fix: vApp instantiation fails if template has more than one VM

shashim22:
There is a bug where vApp instantiation fails if the template has
multiple VMs and any of VMs has more than one NICs. And, any of
NIC is connected to a vDC network. It works fine if VMs are only
connected to vApp networks.

Existing API was designed on an assumption that there will be only
one VM having one NIC. Restructured the code to handle multiple VMs.
Behaviour will remain same for a single VM template. In case of
multiple VMs, customisation parameters will be ignored.

Testing Done:
Ran vapp_tests.py, related tests are passing.
Also did manual testing of following scenarios:
- Single VM template with one NIC and no customisation
- Single VM template with one NIC and network customisation
- Two VMs template with two NICs, one connected to a vDC network
and other to a vApp network

@rocknes @rajeshk2013

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/713)


rajeshk2013 (approved):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/713)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #718, opened 2020-11-04
Update vApp to create VMs from template

antonio-napoles:
-Short description: Makes easier to create a new VM and attach it to the vApp by providing the vapp template name and VM template name as input parameters. This also works by proving a vapp name and VM name to create a VM based on any VM from a different vApp.

-Detailed description: Adding a couple of methods to extend the functionality of vApp. Define method create_vm_from_template, this is an extension from add_vm in order to simplify the VM creation. In this method, a vapp template is retrieved from a catalog to be used as an input parameter for add_vm. But create_vm_from_template facilitates the process by accepting the vapp template name as a str parameter. The other method get_template, is used simply to get the tempalte resource the add_vm methods needs as a paramter.

For me as a developer was complex to figure out a way to create a VM based on a template and attached to my vApp by looking at the defined methods. I guess this method will make this operation more accesible and understandable for consumers.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/718)

### DIFF
--- a/pyvcloud/vcd/vapp.py
+++ b/pyvcloud/vcd/vapp.py
@@ -885,7 +885,6 @@ class VApp(object):
         :param dict spec: a dictionary containing
 
             - vapp: (resource): (required) source vApp or vAppTemplate
-            
                 resource.
             - source_vm_name: (str): (required) source vm name.
             - target_vm_name: (str): (optional) target vm name.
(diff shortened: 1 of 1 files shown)


antonio-napoles:
Tracking this work on a previous branch.

PR #756, opened 2021-03-12
Add a timeout param to _do_request_prim

rocknes:
_do_request_prim in client.py doesn't have a timeout parameter, as a result there is no way to test connectivity to server without waiting for the request to timeout, which can be in minutes. By adding the timeout parameter, the client can control how long they need to wait before receiving a ReadTimeoutError.

Testing done:
Wrote a custom script that made a GET /api/cse/system call to vCD with no CSE server running,
1. without timeout the call took 30+ seconds to fail
2. With a timeout of 3.05 seconds, the call returned the expected error in ~3 seconds.

### DIFF
--- a/pyvcloud/vcd/client.py
+++ b/pyvcloud/vcd/client.py
@@ -1269,7 +1269,8 @@ class Client(object):
                          accept_type=None,
                          auth=None,
                          params=None,
-                         extra_headers=None):
+                         extra_headers=None,
+                         timeout=None):
         headers = extra_headers or {}
         if media_type is not None:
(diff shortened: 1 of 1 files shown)


Anirudh9794 (commented):
:lgtm: 

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/756#-:-MVqkIw0DyHm9DmjSVJ7:b593w57)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @shashim22)

PR #766, opened 2021-06-08
Stop pyvcloud creating an empty log file called 'vcd_sdk.log'

andrew-lee-1089:
Remove the need to call "Get_logger" which inadvertently creates an empty `vcd_sdk.log` file in your current working directory, which seems un-useful for a single log across the whole project.

pyvcloud is great -but there is no logging, I think that is basically fine so long as we are consistent with that fact, implementing logging for one log feels like the worst of both worlds, and is leaving us with an empty file in our current working directory, with no way to configure this log file name or location. I rejected the idea of doing anything fancier here,  because sorting out logging in this project felt too big a task for a Tuesday morning. 

I've not tested. 

Can someone please review?

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/766)

### DIFF
--- a/pyvcloud/vcd/vapp.py
+++ b/pyvcloud/vcd/vapp.py
@@ -24,7 +24,6 @@ from pyvcloud.vcd.client import E_OVF
 from pyvcloud.vcd.client import EntityType
 from pyvcloud.vcd.client import FenceMode
 from pyvcloud.vcd.client import find_link
-from pyvcloud.vcd.client import get_logger
 from pyvcloud.vcd.client import MetadataDomain
 from pyvcloud.vcd.client import MetadataValueType
 from pyvcloud.vcd.client import MetadataVisibility
@@ -42,7 +41,6 @@ from pyvcloud.vcd.vdc import VDC
 from pyvcloud.vcd.vm import VM
 
 DEFAULT_CHUNK_SIZE = 10 * 1024 * 1024
-LOGGER = get_logger()
 
 
 class VApp(object):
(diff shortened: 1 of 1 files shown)


rocknes (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/766#-:-McrdzsrA063wt6P6epl:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @andrew-lee-metaswitch)

___
*[pyvcloud/vcd/vapp.py, line 1605 at r1](https://reviewable.io/reviews/vmware/pyvcloud/766#-Mcrdav09AE-KAZTapc7:-Mcrdav17o0xW_ECZfkk:b6t1sy3) ([raw file](https://github.com/vmware/pyvcloud/blob/51f433d5617bfec142e72885aed7e35cd53162af/pyvcloud/vcd/vapp.py#L1605)):*
> ```Python
>                 no_of_vm_upgraded += 1
>             except OperationNotSupportedException:
>                 raise OperationNotSupportedException(
> ```

Under normal circumstance raising an exception here would make perfect sense. However, by changing the current behavior (no exception -> raising exception) we risk breaking existing users of this method.

We will merge this in the next major release of pyvcloud.

andrew-lee-metaswitch:
If I changed it from logger.error to just print - would that be acceptable / not-deemed a non-breaking change?

andrew-lee-1089:
@rocknes any update on this? I'd like to get this released if possible.

Thanks!

volehuy1998:
@rocknes , same here. Could we sync path log from client initialzation to this or optional it by path?
`def _get_default_logger(self, file_name="vcd_pysdk.log",`

PR #720, opened 2020-11-12
Fix AttributeError in delete_nic

dwettstein:
Fix a possible AttributeError in `vm.py:delete_nic()` method when NetworkConnection is not available.

PR #788, opened 2021-12-15
Add method delete_all_nics to vm.py

dwettstein:
This PR add the method `delete_all_nics` to `vm.py`. The logic is basically the same as in the `delete_nic` method, except that instead of deleting just the NIC with given index, it deletes all.

NOTE: I have not added a dedicated test for this method, because of the same reason as above.

PS: I recommend to merge also PR #720, as it contains some bugfixes for `delete_nic` method.

### DIFF
--- a/pyvcloud/vcd/vm.py
+++ b/pyvcloud/vcd/vm.py
@@ -739,6 +739,27 @@ class VM(object):
             net_conn_section, RelationType.EDIT,
             EntityType.NETWORK_CONNECTION_SECTION.value, net_conn_section)
 
+    def delete_all_nics(self):
+        """Deletes all nics from the VM.
+
+        :return: an object containing EntityType.TASK XML data which represents
+            the asynchronous task deleting all nics.
+
+        :rtype: lxml.objectify.ObjectifiedElement
+        """
+        # get network connection section.
+        net_conn_section = self.get_resource().NetworkConnectionSection
+
+        if hasattr(net_conn_section, 'NetworkConnection'):
+            for nc in net_conn_section.NetworkConnection:
+                net_conn_section.remove(nc)
+        else:
+            raise InvalidStateException(
+                'No nics found in the VM' % (self.get_resource().get('name')))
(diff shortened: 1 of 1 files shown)


rocknes (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/788)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dwettstein and @rocknes)

___
*[pyvcloud/vcd/vm.py, line 757 at r1](https://reviewable.io/reviews/vmware/pyvcloud/788#-MrTtnk91Yyl-1kMcvzN:-MrTtnkAABXpOt_jJIod:bsd059y) ([raw file](https://github.com/vmware/pyvcloud/blob/bbe9a04f46db2759539e008302869d9e1fbb0bec/pyvcloud/vcd/vm.py#L757)):*
> ```Python
>                 net_conn_section.remove(nc)
>         else:
>             raise InvalidStateException(
> ```

I think we don't need to raise an exception if no nics are found.

dwettstein (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/pyvcloud/788)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @rocknes)

___
*[pyvcloud/vcd/vm.py, line 757 at r1](https://reviewable.io/reviews/vmware/pyvcloud/788#-MrTtnk91Yyl-1kMcvzN:-MrYj8Qx4KKfkDh7Z6dp:ba4vq2y) ([raw file](https://github.com/vmware/pyvcloud/blob/bbe9a04f46db2759539e008302869d9e1fbb0bec/pyvcloud/vcd/vm.py#L757)):*

I think we don't need to raise an exception if no nics are found.
</blockquote></details>

Thanks @rocknes, you're right, I have removed the `else` part.

dwettstein:
Hi @rocknes 

I have implemented your requested changes. Is there something left to do for this PR?

Thanks.

